PR #8, opened 2023-11-15
remove unneeded workflow download

yoachim:


rhiannonlynne:
I think due to my inadvertent edit of the workflows on main, that you probably don't need this anymore.

PR #9, opened 2023-11-15
Tickets/opsim 1099

rhiannonlynne:
Update workflows 
* Tried Miniconda, and it's a few minutes slower than mamba forge so sticking with mamba forge for now
* Moved pypi upload into workflow running unit tests (so tested on both Mac and Linux before upload)
* Moved doc build out of workflow with unit tests and into its own workflow 
* Added cache of scheduler data to actions (this will run each night, to update cache)
* Using cache in workflows for unit tests and doc builds. (ubuntu downloads were starting to fail, maybe for whatever mystery reason they were failing on schedview). 

Added --update flag to scheduler_download_data so that updating only directories which are not in-sync with expected build keys will be updated; this lets us use a generic "rubin_sim_data" cache and just update what is necessary on the runner which needs it.
Did some updates to get docs to build better.


rhiannonlynne:
If this all looks good, then a merge to main + tag (v0.0.2.alpha?) should test upload to pypi.

rhiannonlynne:
Docs built at  https://rubin-scheduler.lsst.io/v/OPSIM-1099/index.html

yoachim (approved):
That all looks like it works

PR #10, opened 2023-11-25
U/lynnej/update readme

rhiannonlynne:
Move additional installation information from README into documentation.  (note that readme goes onto pypi, as a typically short description .. see https://pypi.org/project/rubin-scheduler/). 
Add more badges to README.


yoachim (approved):
Looks fine. Could probably squash a few of the commits together.

PR #11, opened 2023-11-27
swap starting_pa for pa typo

yoachim:
Nasty little typo that resulted in some rare long slew times because the kinematic model thought it had to make a large rotation.


ehneilsen (approved):

PR #12, opened 2023-11-27
Tickets/opsim 1102

yoachim:
Update `sim_runner` so it pre-allocates an array to hold observations. Results in a modest speedup, and makes it possible to keep the total memory footprint under 8 GB for a 10 year simulation.


ehneilsen (approved):
Looks fine.
I can imagine circumstances where using a nan might be a problem, for example if there are ever times when there might be an undefined RA, for example in the unlikely event that the telescope needs to track a quickly moving object (obviously not part of the main survey, but something that happens on telescopes). You _could_ use `counter` in line 144 instead.

PR #13, opened 2023-11-28
deprecate thomson and training dirs

yoachim:
Drop two old cruft directories.


rhiannonlynne (approved):
Good idea. Do you think we'd ever want to do something like this in the future, where having these pulled out to say (lsst-sims) and archived in a repo there would make it easier to find?

yoachim:
> Good idea. Do you think we'd ever want to do something like this in the future, where having these pulled out to say (lsst-sims) and archived in a repo there would make it easier to find?

I don't think we need to save these. The optimizer code needs a single merit function to optimize, and the thomson code is not really needed now that we've found the source of our tesselation.

PR #15, opened 2023-12-05
Tickets/opsim 893

yoachim:
Updates to `IntRounded` so behavior is once again repeatable cross-platforms. Have `IntRoudned` throw an error if a NaN is passed in.


yoachim:
> Looks fine, but two questions:
> 
>     * is this the cause of all the warning message that were popping up regarding "can't cast as value" (or something similar) when running schedview,
> 
>     * would it make sense to have the masking of NaN happen inside the IntRounded call, to safeguard existing behavior

* yes, this was throwing lots of "can't cast" warnings
* It would make sense to do the masking inside `IntRounded`, however, ints do not have any sort of NaN value available in python, so the easy solution is to throw an error if a user tries to send one in.

PR #17, opened 2023-12-14
modelobservatory can run all the way to sunrise and set

yoachim:
Let the model observatory keep running all the way to sunrise/sunset before it forces a jump to the next night.

### DIFF
--- a/rubin_scheduler/scheduler/model_observatory/model_observatory.py
+++ b/rubin_scheduler/scheduler/model_observatory/model_observatory.py
@@ -57,6 +57,8 @@ class ModelObservatory:
         downtimes=None,
         no_sky=False,
         wind_data=None,
+        starting_time_key="sun_n12_setting",
+        ending_time_key="sun_n12_rising",
     ):
         """
         Parameters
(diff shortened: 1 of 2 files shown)


rubin_scheduler/scheduler/model_observatory/model_observatory.py
```diff
@@ -486,18 +498,18 @@ def check_mjd(self, mjd, cloud_skip=20.0):
             while clouds > self.cloud_limit:
                 new_mjd = new_mjd + cloud_skip / 60.0 / 24.0
                 clouds = self.cloud_data(Time(new_mjd, format="mjd"))
-        alm_indx = np.searchsorted(self.almanac.sunsets["sunset"], mjd) - 1
+        alm_indx = np.searchsorted(self.almanac.sunsets["sunset"], mjd, side="right") - 1
```
rhiannonlynne, line 501: I'm a little confused about the swap from the general search sorted to "right" side search sorted. 
In particular, if this is the right thing to do here, why isn't it the right thing to do in "Almanac.mjd_index"? 
And earlier in model_observatory, self.almanac_index is defined differently as well  -- and why wouldn't self.almanac_indx be the place to start with, instead of alm_indx?

yoachim: The attribute `self.almanac_index` corresponds with `self.mjd`. This method is checking an arbitrary mjd, so it will potentially be looking at a different spot in the almanac. 

The `side="right"` is necessary for the edge case of an mjd being equal to a value in the almanac. Never happened before because we always skipped over sunrise and sunset. Otherwise it can fail to skip forward to a new night. Probably should be updated in `Almanac.mjd_index` as well.

rhiannonlynne (approved):

PR #21, opened 2024-01-11
Tickets/opsim 1104

yoachim:
Adding a new `Survey` sub-class for cases where one has a list of pointings to chose from. 

Examples using the new class over in `rubin_sim_notebooks`


tests/scheduler/test_baseline.py
```diff
@@ -11,12 +11,52 @@
...
+    GreedySurvey,
+    ScriptedSurvey,
+    generate_ddf_scheduled_obs,
+)
 from rubin_scheduler.scheduler.utils import SkyAreaGenerator, calc_norm_factor_array
 
 SAMPLE_BIG_DATA_FILE = os.path.join(get_data_dir(), "scheduler/dust_maps/dust_nside_32.npz")
 
 
+class ModelObservatoryWindy(ModelObservatory):
```
ehneilsen, line 25: This seems like something that would be better handled by customizing ModelObservatory rather than a whole new subclass?

ehneilsen (commented):
Mostly looks okay, except for the need for more detailed docstrings.

tests/scheduler/test_baseline.py
yoachim: I did it as a subclass because I don't think there's a normal use case where we would want a model observatory where it's always windy. Didn't want to add a kwarg that could only ever be used to break things in production.

rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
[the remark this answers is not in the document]
yoachim: Yeah, I think I like that better too. I'll change the H.A. in the conditions object to default to -pi to pi. I'll do that on a new branch because there are going to be a number of places where that comes up.

ehneilsen (approved):

PR #22, opened 2024-01-13
Tickets/opsim 1111

yoachim:
Add altitude and azimuth limits to the conditions object. Add new basis function that can grab and interpret the limits, as well as expand the masks in case a survey will be taking observations in the future.

### DIFF
--- a/rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
+++ b/rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
@@ -12,7 +12,7 @@ __all__ = (
     "AltAzShadowMaskBasisFunction",
 )
 
-from warnings import warn
+import warnings
 
 import healpy as hp
 import matplotlib.pylab as plt
@@ -304,8 +304,9 @@ class ZenithShadowMaskBasisFunction(BaseBasisFunction):
         penalty=np.nan,
         site="LSST",
     ):
-        warn.DeprecationWarning(
-            "Deprecating ZenithShadowMaskBasisFunction in favor of AltAzShadowMaskBasisFunction."
+        warnings.warn(
+            "Deprecating ZenithShadowMaskBasisFunction in favor of AltAzShadowMaskBasisFunction.",
+            DeprecationWarning,
         )
 
         super(ZenithShadowMaskBasisFunction, self).__init__(nside=nside)
--- a/rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
+++ b/rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
@@ -9,6 +9,7 @@ __all__ = (
     "MaskAzimuthBasisFunction",
     "SolarElongationMaskBasisFunction",
     "AreaCheckMaskBasisFunction",
+    "AltAzShadowMaskBasisFunction",
 )
 
 import healpy as hp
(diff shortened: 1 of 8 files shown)


rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
```diff
@@ -194,6 +195,77 @@ def _calc_value(self, conditions, indx=None):
         return result
 
 
+class AltAzShadowMaskBasisFunction(BaseBasisFunction):
+    """Mask any out of range altitudes and azimuths, then extend the
+    mask so if observations are taken in pairs, the second in the pair will
+    not have moved into a masked region.
+    """
```
rhiannonlynne, line 202: Is this intended to be used as the basic "mask out of range values" or only in the case of "extend the mask for out of range values to include a shadow region" ? (I think it's both, and the shadow could be 0, but the docstring could be more clear. Also it needs to include the parameters and explain what shadow_minutes means.) 

Think of it as explaining the usage and purpose of this class to people who mostly know what the scheduler does, but don't know anything about the details and shouldn't have to read the code to find out.

rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
```diff
@@ -194,6 +195,77 @@ def _calc_value(self, conditions, indx=None):
...
+        self.shadow_time = shadow_minutes / 60.0 / 24.0  # To days
+        self.result = np.zeros(hp.nside2npix(self.nside), dtype=float)
+        self.in_range = np.zeros(hp.nside2npix(self.nside), dtype=int)
+
+    def _calc_value(self, conditions, indx=None):
+        # Mask everything to start
+        result = self.result.copy() + np.nan
+
+        in_range_alt = self.in_range.copy()
+        in_range_az = self.in_range.copy()
```
rhiannonlynne, line 223: Why do these need to be copies? Is it so that the original 'result' and 'in_range' remain at 0 instead of being modified and then re-modified? 
Is it better to make the copies or is it better to just set result=np.zeros etc. each time _calc_value is called? (I think you may have tested this?).

rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
rhiannonlynne: Oh yeah, and I guess you want to also explain what the conditions.alt_limits and conditions.az_limits do here too and how they interact with min_alt and max_alt.

rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
```diff
@@ -194,6 +195,77 @@ def _calc_value(self, conditions, indx=None):
...
+        result[np.where(IntRounded(conditions.alt) < IntRounded(self.min_alt))] = np.nan
+        result[np.where(IntRounded(conditions.alt) > IntRounded(self.max_alt))] = np.nan
+
+        result[np.where(IntRounded(future_alt) < IntRounded(self.min_alt))] = np.nan
+        result[np.where(IntRounded(future_alt) > IntRounded(self.max_alt))] = np.nan
+
+        return result
+
+
 class ZenithShadowMaskBasisFunction(BaseBasisFunction):
```
rhiannonlynne, line 269: Should we deprecate this shadow mask function? (need to deprecate ahead of actually removing it).

rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
```diff
@@ -194,6 +195,77 @@ def _calc_value(self, conditions, indx=None):
...
+        in_range_az = self.in_range.copy()
+
+        # Compute the alt,az values in the future. Use the conditions object
+        # so the results are cached and can be used by other surveys is needed.
+        # Technically this could fail if the masked region is very narrow or shadow time
+        # is very large.
+        future_alt, future_az = conditions.future_alt_az(np.max(conditions.mjd + self.shadow_time))
+
+        # apply limits from the conditions object
+        for limits in conditions.alt_limits:
```
rhiannonlynne, line 232: I'm confused, because I don't see an "alt_limits" attribute in conditions.

rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
yoachim: Yup, just resetting them to zero each time. The speed looks to be identical, so might as well call `np.zeros` and save a few lines.

rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
yoachim: It's a little more clever than the new mask, but probably not enough to keep it around.

rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
yoachim: ok, got that sorted so everthing uses `tel_alt_limits` properly now.

rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
rhiannonlynne: Improving this docstring would be really helpful for people using this function, particularly to know the interaction with the conditions values.

rhiannonlynne (commented):
There is a unit test failing because of the new DeprecationWarning. 
Please extend the documentation for the AltAzShadowMaskBasisFunction to include information about how it uses the conditions alt/az limits.

rhiannonlynne (approved):

rhiannonlynne:
Looks good. This is a nice feature to add, especially for testing use.

PR #24, opened 2024-01-18
tickets/PREOPS-4782: when requesting time during the day, set available block size time to 0

ehneilsen:


rhiannonlynne (approved):
Pretty sure this was Peter's recommendation and it seems to work. 
You found this error was triggered by something, and if not too extensive, it would be good to add that unit test -- but I'm not sure how far you had to go to find the edge case.

ehneilsen:
> Pretty sure this was Peter's recommendation and it seems to work. You found this error was triggered by something, and if not too extensive, it would be good to add that unit test -- but I'm not sure how far you had to go to find the edge case.

Okay, added a test.

PR #27, opened 2024-01-18
tickets/PREOPS-4781: make save of scheduler pickle work for arbitrary paths, not just from the current dir.

ehneilsen:


yoachim (approved):
yup

PR #28, opened 2024-01-20
let scripted observations check for twilight starting

yoachim:
Update scripted surveys to have logic that will prevent long sequences from executing right before twilight. Mainly for DDFs.

### DIFF
--- a/rubin_scheduler/scheduler/surveys/scripted_surveys.py
+++ b/rubin_scheduler/scheduler/surveys/scripted_surveys.py
@@ -28,6 +28,8 @@ class ScriptedSurvey(BaseSurvey):
     return_n_limit : `int` (10)
         The maximum number of observations to return. Set to high and your block
         of scheduled observations can run into twilight time.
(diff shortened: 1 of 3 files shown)


rubin_scheduler/scheduler/surveys/scripted_surveys.py
```diff
@@ -267,7 +272,28 @@ def set_script(self, obs_wanted):
...
-        return observations
+        observations = self._check_list(conditions)
+        if observations is None:
+            self.observations = []
+            self.last_mjd = conditions.mjd
+            return self.observations
+
+        # If we want to ensure the observations can be completed before twilight starts
+        if self.before_twi_check:
+            # not bothering to include filter changes and read time now.
```
ehneilsen, line 287: Maybe you should include them, though? Particularly for ToO followups (which might be implemented with scripted surveys), they might ask for lots of filter changes, which could be significant.

ehneilsen (approved):
Looks fine.

PR #29, opened 2024-01-22
tickets/PREOPS-4795: Use SITCOMTN-32 dayObs defn. for rollover time in prenight sim archive

ehneilsen:


rhiannonlynne (approved):
I think this is a good idea, helps when everyone is using the same language.

PR #30, opened 2024-01-23
OPSIM-1113: bugfix for dither detailer (wrap RA) and DDF prescheduling for morning twilight

yoachim:
Make sure dither detailer properly wraps coordinates so we don't observe negative RAs. 

Update the DDF pre-scheduling to make sure a long sequence is not scheduled right before twilight starts.


ehneilsen (approved):
Looks good to me.

PR #32, opened 2024-01-25
Tickets/opsim 1114 Add jerk to the telescope kinematic model

yoachim:
Update the kinematic model of the telescope to include jerk. All jerks and accelerations assumed to be symmetric (e.g., maximum deceleration equals negative maximum acceleration). Removed previously unused deceleration kwargs.

### DIFF
--- a/rubin_scheduler/scheduler/model_observatory/jerk.py
+++ b/rubin_scheduler/scheduler/model_observatory/jerk.py
@@ -20,7 +20,7 @@ def jerk_time(distance, v_max, acc_max, jerk_max):
     acc_max : `float`
         The maximum acceleration. Units of length or angle per unit of time squared.
     jerk_max : `float`
-        The maximum jerk. Units of length or angle per unit of time cubed. If 
+        The maximum jerk. Units of length or angle per unit of time cubed. If
         set to None, assumes jerk is infinite and defaults to use acc_time.
     """
 
@@ -31,14 +31,12 @@ def jerk_time(distance, v_max, acc_max, jerk_max):
     if jerk_max is None:
         return acc_time(distance, v_max, acc_max)
 
-    trajectory_instance_case = get_trajectory_instance_case(
-        distance, v_max, acc_max, jerk_max
-    )
+    trajectory_instance_case = _get_trajectory_instance_case(distance, v_max, acc_max, jerk_max)
 
     unique_trag = np.unique(trajectory_instance_case)
 
     if np.size(unique_trag) == 1:
-        tj, ta, tv = calculate_times(
+        tj, ta, tv = _calculate_times(
             distance,
             v_max,
             acc_max,
@@ -50,7 +48,7 @@ def jerk_time(distance, v_max, acc_max, jerk_max):
         result = np.zeros(np.size(distance))
         for tic in unique_trag:
             indx = np.where(trajectory_instance_case == tic)[0]
-            tj, ta, tv = calculate_times(
+            tj, ta, tv = _calculate_times(
                 distance[indx],
                 v_max,
                 acc_max,
@@ -88,7 +86,11 @@ def acc_time(distance, v_max, acc_max):
     return slew_time
 
 
-def calculate_times(distance, v_max, acc_max, jerk_max, trajectory_instance_case=1):
+def _calculate_times(distance, v_max, acc_max, jerk_max, trajectory_instance_case=1):
+    """Calculate travel time including jerk.
+
+    Modified from https://github.com/mdhom/py_constant_jerk/blob/main/constantJerk.py
+    """
     if trajectory_instance_case == 1 or trajectory_instance_case == 3:
         tj = np.sqrt(v_max / jerk_max)
         ta = tj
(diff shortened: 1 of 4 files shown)


rhiannonlynne:
I tried this out. 
Good - The new KinemModel does exactly match the old KinemModel when no jerk is added (I know this was a case of "it should definitely do so, but I figured it was worth checking). 
Bad - Then I tried to add some jerk, but didn't know what values to use, so .. since v_max and a_max are the same values for the telescope in the same direction (alt or az) .. I figured what the heck and stuck in a jerk of the same value as well.  
This led to a failure of the code -- I was getting trajectory instances of 0. 

So then of course, I was kind of confused and so I went and looked at the code and the repository where the code originally came from (they do match up), and I also looked at the original paper being cited. The original paper is also summarized here - https://www.jpe-innovations.com/precision-point/third-order-point-to-point-motion-profile/#:~:text=In%20many%20mechatronic%20applications%20where,speed%2C%20acceleration%20and%20jerk). 
I think the basic problem is just that ">=" probably should be substituted where ">" alone was used sometimes. 
(I will note the definitions of t_j, t_a, t_v are kind of confusing in themselves, because I would have thought "t_a" was the part where you were either coasting under constant acceleration or included both acceleration and deceleration, but it seems like it's neither .. so t_j plus t_a plus t_v does indeed make the total slew time, instead of being half of the total slew time). 

Note: we should probably make sure to include a reference to this link as well as the GitHub repo where you found the code: 
https://www.researchgate.net/publication/289374755_THIRD_ORDER_POINT-TO-POINT_MOTION_-PROFILE
regarding the source of the underlying equations. 

Anyway I did find that making the following modifications allowed the calculation to work, even when using the jerk values I coincidentally chose to be the absolute worst for the first experiment: 

```
def _get_trajectory_instance_case(distance, v_max, acc_max, jerk_max):
    """Determine which motion profile to use.

    Modified from https://github.com/mdhom/py_constant_jerk/blob/main/constantJerk.py

    Parameters
    ----------
    distance : `float`
        The distance to travel. Could be units of length or angle as long as other
        parameters match.
    v_max : `float`
        The maximum velocity. Units of length or angle per unit of time.
    acc_max : `float`
        The maximum acceleration. Units of length or angle per unit of time squared.
    jerk_max : `float`
        The maximum jerk. Units of length or angle per unit of time cubed.

    """
    # Which case applies depends on v_max and distance moved.
    # Case 1 & 3 & 4 - v_max < v_a (v_max reached)
    #   [2 & 5 & 6 v_max >= v_a]
    # Case 1 & 5 & 6 - s >= s_a (max acceleration reached?)
    #   [2 & 3 & 4, s < s_a]
    # Case 3 & 5 - s >= s_v (max velocity reached)
    #   [ 4 & 6 - s < s_v]  .. 1 & 2 distance not relevant.
    # Case 1: a_max reached, v_max reached and constant for a time
    #          (s = 100,   j = 2000, a = 500,  vMax = 120)
    # Case 2: a_max not reached, v_max not reached
    #          (s = 15000, j = 2000, a = 5500, vMax = 20500)
    # Case 3: a_max not reached, v_max reached and constant for a time
    #          (s = 15000, j = 2000, a = 5500, vMax = 2500)
    # Case 4: a_max reached, v_max not reached
    #          (s = 57,    j = 2000, a = 500,  vMax = 120)
    # Case 5: a_max reached *and constant for a time*,
    # v_max reached and constant for a time
    #          (s = 15000, j = 2000, a = 500,  vMax = 2500)
    # Case 6: a_max reached and constant for a time, v_max not reached
    #          (s = 15000, j = 2000, a = 500,  vMax = 20500)

    result = np.array(distance * 0)

    # Hypothetical max velocity given acceleration and jerk
    v_a = acc_max * acc_max / jerk_max
    # Hypothetical max distance traveled given acceleration and jerk
    s_a = 2 * acc_max * acc_max * acc_max / (jerk_max * jerk_max)
    # Hypothetical max distance traveled given actual v_max cap
    # if v_max < v_a
    if v_max * jerk_max < acc_max * acc_max:
        s_v = v_max * 2 * np.sqrt(v_max / jerk_max)
    # else if v_max >= v_a
    else:
        s_v = v_max * (v_max / acc_max + acc_max / jerk_max)

    result[np.where((v_max < v_a) & (distance >= s_a))] = 1
    result[np.where((v_max >= v_a) & (distance < s_a))] = 2
    result[np.where((v_max < v_a) & (distance < s_a) & (distance >= s_v))] = 3
    result[np.where((v_max < v_a) & (distance < s_a) & (distance < s_v))] = 4
    result[np.where((v_max >= v_a) & (distance >= s_a) & (distance >= s_v))] = 5
    result[np.where((v_max >= v_a) & (distance >= s_a) & (distance < s_v))] = 6

    return result
```

rhiannonlynne (changes requested):
See needed modifications to case determination function to support the unique case of jerk being equal to max acceleration and max velocity.

rubin_scheduler/scheduler/model_observatory/jerk.py
```diff
@@ -125,6 +127,8 @@ def _get_trajectory_instance_case(distance, v_max, acc_max, jerk_max):
     """Determine which motion profile to use.
 
     Modified from https://github.com/mdhom/py_constant_jerk/blob/main/constantJerk.py
+    which was probably derived from:
+    https://www.researchgate.net/publication/289374755_THIRD_ORDER_POINT-TO-POINT_MOTION_-PROFILE
```
rhiannonlynne, line 131: Did you see  mdhom's other repo here https://github.com/mdhom/PointToPointMotion ? That one is basically the C# version of the same thing, and includes the reference to the paper.

rhiannonlynne (approved):

rubin_scheduler/scheduler/model_observatory/jerk.py
yoachim: Ah, I didn't see that one. I did put in a PR to the original python repo so see if they agree with the change.

PR #34, opened 2024-01-25
tickets/PREOPS-4736: Support loading of sky brightess with lsst.resources in rubin_scheduler

ehneilsen:
This enables loading sky brightness files directly from the http directory, or (if/when there eventually is one) an s3 bucket.
This is intended for use in the containers running dashboards at the USDF.

### DIFF
--- a/rubin_scheduler/skybrightness_pre/sky_model_pre.py
+++ b/rubin_scheduler/skybrightness_pre/sky_model_pre.py
@@ -1,15 +1,25 @@
 __all__ = ("SkyModelPre", "interp_angle")
 
+import abc
 import glob
 import os
+import urllib
 import warnings
+from pathlib import Path
 
 import h5py
 import healpy as hp
 import numpy as np
+import requests
 
+import rubin_scheduler.data.rs_download_sky
 from rubin_scheduler.data import get_data_dir
 from rubin_scheduler.utils import _angular_separation, _hpid2_ra_dec, survey_start_mjd
+
+try:
+    from lsst.resources import ResourcePath
+except ImportError:
+    pass
 
 
 def short_angle_dist(a0, a1):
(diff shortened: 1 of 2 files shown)


timj:
I think the coverage report is broken because it has `--cov=rubin_scheduler` when it needs to be the equivalent of `--cov=lsst.daf.butler --cov=tests`

rubin_scheduler/skybrightness_pre/sky_model_pre.py
```diff
@@ -328,3 +325,100 @@ def return_mags(
...
+        return h5
+
+
+class SkyModelPreWithResources(SkyModelPreBase):
+    def _init_files(self):
+        data_resource = ResourcePath(self.data_path, forceDirectory=True)
+        self.files = []
+
+        try:
+            for dir_path, dir_names, file_names in data_resource.walk(file_filter=r".*\.h5"):
```
timj, line 377: Consider using:
```
self.files = list(ResourcePath.findFileResources([data_resource], file_filter=r".*\.h5"))
```

rhiannonlynne:
Comment/question -- does this change mean that lsst.resource is *required* for skybrightness_pre data loading, or does it mean it's an option? 
I'm really asking whether lsst.resources is a still an optional or if it's become a required dependency after this PR (since it matters for @tribeiro and summit installation, and we should endeavor to not make it required until there is a conda package for lsst.resources).

timj:
If it becomes critical that resources is on conda-forge there is no problem with that happening. It's simply never been a priority.

ehneilsen:
> does this change mean that lsst.resource is _required_ for skybrightness_pre data loading, or does it mean it's an option?

Currently, it's optional if you only want to use local sky brightness files, but if you want to use remote copies (e.g. because you are running in a container you can't or don't want to fill with sky brightess files) you need lsst.resources.

I'd rather not maintain the two separate subclasses, though, so I'll be happy when we can just use lsst.resources as a dependency.

rhiannonlynne (approved):
I believe Peter had a look at this and commented in slack that it was fine.

PR #35, opened 2024-01-26
rotator documentation

yoachim:
Expand documentation on the rotTelPos and rotSkyPos options within the observation object.

### DIFF
--- a/rubin_scheduler/scheduler/utils/utils.py
+++ b/rubin_scheduler/scheduler/utils/utils.py
@@ -481,12 +481,18 @@ def empty_observation(n=1):
     filter : `str`
         The filter used. Should be one of u, g, r, i, z, y.
     rotSkyPos : `float`
-        The rotation angle of the camera relative to the sky E of N (Radians). Will probably be overridden if rotTelPos is not np.nan.
+        The rotation angle of the camera relative to the sky E of N (Radians). Will be ignored if rotTelPos is finite.
+        If rotSkyPos is set to NaN, rotSkyPos_desired is used.
     rotTelPos : `float`
         The rotation angle of the camera relative to the telescope (radians). Set to np.nan to force rotSkyPos to be used.
     rotSkyPos_desired : `float`
-        If both rotSkyPos and rotTelPos are None/NaN, then rotSkyPos_desired is used. If rotSkyPos_desired results in a valid
-        rotTelPos, rotSkyPos is set to rotSkyPos_desired. Otherwise, something else will happen--XXX.
+        If both rotSkyPos and rotTelPos are None/NaN, then rotSkyPos_desired (radians) is used.
+        If rotSkyPos_desired results in a valid rotTelPos, rotSkyPos is set to rotSkyPos_desired.
+        If rotSkyPos and rotTelPos are both NaN, and rotSkyPos_desired results in an out of range value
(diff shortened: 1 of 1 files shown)


rubin_scheduler/scheduler/utils/utils.py
```diff
@@ -481,12 +481,18 @@ def empty_observation(n=1):
...
     rotSkyPos_desired : `float`
-        If both rotSkyPos and rotTelPos are None/NaN, then rotSkyPos_desired is used. If rotSkyPos_desired results in a valid
-        rotTelPos, rotSkyPos is set to rotSkyPos_desired. Otherwise, something else will happen--XXX.
+        If both rotSkyPos and rotTelPos are None/NaN, then rotSkyPos_desired (radians) is used.
+        If rotSkyPos_desired results in a valid rotTelPos, rotSkyPos is set to rotSkyPos_desired.
+        If rotSkyPos and rotTelPos are both NaN, and rotSkyPos_desired results in an out of range value
+        for the camera rotator, then rotTelPos_backup is used.
+    rotTelPos_backup : `float`
+        Rotation angle of the camera relative to the telescope (radians). Only used as a last resort if 
+        rotSkyPos and rotTelPos are set to NaN and rotSkyPos_desired results in an out of range rotator value.
```
rhiannonlynne, line 495: Am I interpreting this correctly (in terms of order of priority)
"rotTelPos > rotSkyPos > rotSkyPos_desired > rotTelPos_backup" 
(where if any of the above is either np.nan, the next lower priority value is used.)
I'm going to assume what is actually sent to the schedulerCSC is the rotTelPos, but I'm not sure (might be good to clarify too .. if all of these are sent, then it should be clarified that this logic to choose between these is done at the SchedulerCSC (or sim_runner ModelObservatory) level.). 
If the resulting request for rotator angle is out of range when the observation is actually to be acquired, then the observation is rejected (by the schedulerCSC or sim_runner's ModelObservatory). 

I suppose when I put it in the order like above, it's not obvious why rotSkyPos and rotSkyPos_desired would be different.

yoachim: You have the order correct. The addition is that if rotSkyPos_desired is out of range, then it goes on to rotTelPos_backup (it never really makes sense to set rotSkyPos_desired to NaN).

For the sims, the `ModelObservatory` executes the angle selection logic and then sends a `rotTelPos` to the kinematic model for execution. Which technically means the rotSkyPos delivered could be off by an arcmin of what was specified because it takes time to slew to the location, but I don't think anyone cares at that level. 

rotSkyPos is what you set if you absolutely want that rotSkyPos value and the visit should fail if it's not possible, while rotSkyPos_desired is what to use if you don't want the visit to fail just because of the rotator angle.

rhiannonlynne: Ok, that makes sense why you'd have rotSkyPos vs. rotSkyPos_desired as separate items. 
Documentation can have a "Notes" (Note?) section which might be a good place to explain the relationship between these? 
(I guess if I was really being picky, I'd say they should go in the order above in terms of kwargs provided, so that in the doc strings, they also follow in the order of priority .. which might mean that you don't even need the separate Notes bit).

rhiannonlynne (approved):

PR #36, opened 2024-01-26
tickets/PREOPS-4831: fix code coverage tests in github action

ehneilsen:


.github/workflows/test_and_build.yaml
```diff
@@ -75,7 +75,7 @@ jobs:
         run: |
           export RUBIN_SIM_DATA_DIR=~/rubin_sim_data
           #pytest -r a -v 
-          pytest -r a -v --cov=rubin_scheduler --cov-report=xml --cov-report=term --cov-branch
+          pytest -r a -v --cov=rubin_scheduler --cov=tests --cov-report=xml --cov-report=term --cov-branch
```
timj, line 78: I'm a bit confused. Yes, the tests need to be included but how come it was 0% before? That means it's not reporting on rubin_scheduler.

ehneilsen: That wasn't the only issue.
On my devel checkout, I initially wasn't able to reproduce the problem.
But, carefully going through the test_and_build.yaml and making a new environment, I was able to reproduce it by including the `--no-deps` in the `python -m pip install .`. I don't know why it makes a difference: when I don't use it when I create a new environment, it doesn't actually add any dependencies. But, something is clearly different.

timj: Great. I'm pleased you tracked it down. 

Regarding `--no-deps`, you'd want to do a `pip list` after you install your dependencies and a `pip list` after you installed the package with it allowed to use deps -- this might tell you there is some inconsistency between your pyproject.toml and requirements.txt.

ehneilsen: Ah, I think I might understand... the `--no-deps` might not be the only difference between the github action and the environment I used that actually worked. I think I also used pip's `-e` option in the one that worked, which changes where the files actually are, which might make it possible for the coverage measurement to actually recognize them for what they are.
If this is the case, then the changes in this pull request so far are not enough.

timj: I see now that you are getting wonderful test coverage for `tests` and nothing for `rubin_scheduler` 😄

rhiannonlynne: Ohhhh ... "where the code is" is probably exactly the issue. 
We *should* be installing with `pip install -e . --no-deps` because we're using conda to do the dependency installation and not pip installing everything. We might even need `--no-build-isolation` (or whatever the proper way to word that flag is, I just don't exactly understand the requirement, but I see Matt Becker recommending it when you pip install into a conda environment, which is what we're doing here). 

So yes, I bet the problem is that with tests listed explicitly, it's getting the test coverage (what I don't understand is why you have test coverage for unit tests? I must be misunderstanding that), but it must be not knowing where the code is/should be. Even though that --cov=rubin_scheduler *ought* to be pointing it to the right place according to docs I've seen.??

rhiannonlynne: Ah, but I think I might have dropped the -e option out of the install at some point because there were complaints in the workflow ..  but I can't remember exactly what the failure was.

timj: It might be that it's getting confused between the installed package rubin_scheduler and the local directory rubin_scheduler. With `-e` they are the same but without that they are different and it will likely be looking for coverage in the local directory but importing the installed code. For pipelines packages this doesn't trigger because all the code is in a python subdirectory.

ehneilsen: That's my current guess. If so, then this current run should show coverage accurately. If it does, I'll go back and put the --no-deps back and verify that it still works.

rhiannonlynne: I was thinking that with --no-deps it should at least fail in the same way it's failing without --no-deps.

rhiannonlynne: This looks like it's working (https://app.codecov.io/github/lsst/rubin_scheduler/tree/tickets%2FPREOPS-4831) but I admit to being a little confused about what test coverage of the tests means. Should the `--cov tests` be included in the code coverage evaluation or not?

rhiannonlynne (approved):

.github/workflows/test_and_build.yaml
timj: Including the tests is entirely up to you. In science pipelines at times we have had test files where some tests were disabled and we hadn't realized until we saw the strangely low coverage for the files.

PR #37, opened 2024-01-29
tickets/PREOPS-4836: make conda package optional in rubin_scheduler

ehneilsen:
Because we want to be able to pip install rubin_scheduler, and conda isn't a pip package.


yoachim (approved):
Looks good

PR #38, opened 2024-01-30
tickets/PREOPS-4846:  update rubin_scheduler for pytest 8 and black 24

ehneilsen:
See discussion [here](https://lsstc.slack.com/archives/G6MP1HTDW/p1706547380936619).

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,5 +1,5 @@
-pytest<8.0.0
+pytest
 pytest-cov
 pytest-black
-black>=23.0.0,<24.0.0
+black
 ruff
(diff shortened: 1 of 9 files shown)


test-requirements.txt
```diff
@@ -1,5 +1,5 @@
-pytest<8.0.0
+pytest
 pytest-cov
 pytest-black
```
rhiannonlynne, line 3: Does this need to remove pytest-black here too?

rhiannonlynne (approved):
pytest-black may need to come out of test-requirements too

ehneilsen:
> pytest-black may need to come out of test-requirements too

need to? No, we don't call it, and just installing it and ignoring it doesn't cause problems. But, there's no need for it to be there, so I'll put in a PR to remove it.

PR #42, opened 2024-02-13
Update ComCam

yoachim:
Fix bit-rotted comcam option, include a comcam unit test.

Notebook with ComCam example forthcoming.

### DIFF
--- a/rubin_scheduler/scheduler/surveys/surveys.py
+++ b/rubin_scheduler/scheduler/surveys/surveys.py
@@ -8,7 +8,6 @@ import numpy as np
 
 from rubin_scheduler.scheduler.surveys import BaseMarkovSurvey
 from rubin_scheduler.scheduler.utils import (
-    HpInComcamFov,
     empty_observation,
     gnomonic_project_toxy,
     int_binned_stat,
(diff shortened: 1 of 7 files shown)


rubin_scheduler/scheduler/surveys/surveys.py
```diff
@@ -382,7 +382,7 @@ def simple_order_sort(self):
 
         # Assuming reward has already been calcualted
 
-        potential_hp = np.where(~np.isnan(self.reward) == True)
+        potential_hp = np.where(np.isfinite(self.reward) == True)[0]
```
ehneilsen, line 385: Is the `== True` really necessary?

rubin_scheduler/scheduler/utils/utils.py
```diff
@@ -839,30 +842,37 @@ def __init__(self, nside=None, side_length=0.7):
             ]
         )
 
-    def __call__(self, ra, dec, rot_sky_pos=0.0):
+    def __call__(self, ra, dec, rotSkyPos=0.0):
```
ehneilsen, line 845: Isn't the underscore form what we actually want?

ehneilsen (approved):
Looks fine, with just a couple points of confusion.

rubin_scheduler/scheduler/utils/utils.py
yoachim: yeah, `rotSkyPos` and `rotTelPos` are the two things that managed to get grandfathered in and not updated to proper snake_case because that's how they are in the `observation` array and the output database schema.

PR #46, opened 2024-02-27
Add 'if_exists' flag when writing to opsim database, for use at summit.

rhiannonlynne:
When using the scheduler at the summit, Tiago uses SchemaConverter to write incrementally to a sqlite file. This change supports this use.


rhiannonlynne:
Ah - I just noticed that this overlaps with your existing PR. 
Could you take the docstrings that Tiago so helpfully added to his version of the update (on ts_scheduler) and add them to the SchemaConverter here? 
And then could close the other PR, since you have the same thing here along with additional unit tests.

yoachim:
Change going in on tickets/OPSIM-1127

PR #48, opened 2024-03-05
add mjd_start to calc_season

yoachim:
Season zero-point needs to be defined by a time and RA to be repeatable for multiple calls. 

Also moved some example commented code into a real function.

### DIFF
--- a/rubin_scheduler/utils/season_utils.py
+++ b/rubin_scheduler/utils/season_utils.py
@@ -22,12 +22,6 @@ def calc_season(ra, mjd):
         The RA (in degrees) of the point on the sky
     mjd : `np.ndarray`
         The times of the observations, in MJD days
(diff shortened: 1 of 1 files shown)


rubin_scheduler/utils/season_utils.py
```diff
@@ -16,14 +18,23 @@ def calc_season(ra, time):
     ----------
     ra : `float`
         The RA (in degrees) of the point on the sky
-    time : `np.ndarray`
+    mjd : `np.ndarray`
         The times of the observations, in MJD days
+    mjd_start : `float`
```
rhiannonlynne, line 23: Docstring should match call signature

rhiannonlynne (changes requested):
Good addition to force RA to be a single number. 
The docstring should match the actual function call though.

PR #49, opened 2024-03-06
OPSIM-1133: Docstrings and comments line lengths to pass ruff

yoachim:
Fix all the docstring and comment lengths so ruff passes. Should be no code touched.


rhiannonlynne (approved):
Indeed, this just changed docstring and comment line length. I did check that the docstrings continue to render in the API docs (default parameters are not intended to be included in the parameter type definition line, return values should be named, backticks should be used for type definition, etc), but it should be the same as previous.

PR #51, opened 2024-03-07
move IntRounded

yoachim:
Snippet of code that was supposed to be helping with cross-platform repeatability, but there was a sneaky square root converting ints to floats, so removed some unneeded float to ints.


ehneilsen (approved):
Looks fine.

PR #52, opened 2024-03-09
OPSIM-1135 Move 'get_baseline' to rubin-sim

rhiannonlynne:
Moves get_baseline to rubin_sim (only). 
The sim baseline is only downloaded with rs_download_data, so having the call to get_baseline() in rubin-scheduler is confusing.

Also updates workflows for node.js 20.


yoachim (approved):
Seems fine.

PR #54, opened 2024-03-14
Include a moon distance limit for scheduled observations.

yoachim:
Include a moon distance limit for scheduled observations. Prevents observations that are delayed a few days from having the moon sneak up on them.


rhiannonlynne (approved):
Looks good.

PR #63, opened 2024-05-07
OPSIM-1151: return a bright sky map if time requested when sun is up

yoachim:
Update `SkybrightnessPre` so it returns a "bright" sky map rather than the nearest twilight map if a time during the day is requested.

### DIFF
--- a/rubin_scheduler/skybrightness_pre/sky_model_pre.py
+++ b/rubin_scheduler/skybrightness_pre/sky_model_pre.py
@@ -11,10 +11,13 @@ import h5py
 import healpy as hp
 import numpy as np
 import requests
+from astropy import units as u
+from astropy.coordinates import AltAz, EarthLocation, SkyCoord, get_sun
+from astropy.time import Time
 
 import rubin_scheduler.data.rs_download_sky
 from rubin_scheduler.data import get_data_dir
-from rubin_scheduler.utils import _angular_separation, _hpid2_ra_dec, survey_start_mjd
+from rubin_scheduler.utils import Site, _angular_separation, _hpid2_ra_dec, survey_start_mjd
 
 try:
     from lsst.resources import ResourcePath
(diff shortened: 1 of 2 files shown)


rubin_scheduler/skybrightness_pre/sky_model_pre.py
```diff
@@ -73,6 +76,34 @@ def interp_angle(x_out, xp, anglep, degrees=False):
...
+        Name of the filter, default "r". Currently unused, but
+        could be useful in the future if a more complicated function
+        gets subbed in.
+    bright_val : `float`
+        The value to plug into the sky. Default 2 mag/sq arcsec.
+    sky_alt_min : `float`
+        Set all sky below the alt limit to NaN. Default 20 degrees.
+    """
+
+    result = sky_alt * 0 + np.nan
```
ehneilsen, line 102: `np.full_like(sky_alt, np.nan)` might make it even more obvious what you're up to.

rubin_scheduler/skybrightness_pre/sky_model_pre.py
```diff
@@ -280,14 +327,36 @@ def return_mags(
...
+                    sbs[filter_name] = simple_daytime(
+                        hp_aa.alt.deg,
+                        hp_aa.az.deg,
+                        sun_alt_az.alt.deg,
+                        sun_alt_az.az.deg,
+                        filter_name=filter_name,
+                    )
+
+            else:
+                warnings.warn("Requested MJD between sunrise and sunset, returning closest maps")
```
ehneilsen, line 352: This looks fine, but a next level more sophisticated (but still pretty poor) would be to keep the map in alt/az space rather than ra/dec space, so part of the sky covered by the map doesn't slip off of the zenith distance range accessible to the telescope.

ehneilsen (approved):

PR #64, opened 2024-05-13
OPSIM-1150: update rotSkyPos / rotTelPos conversion for rubin fov

yoachim:
Update rotTelPos and rotSkyPos relatio, making it easier to have different relations for the main tel and auxtel.


rhiannonlynne:
I'm not sure I understand what rotSkyPos and rotTelPos is being passed out with the observation in an observation request. I see the new rotSkyPos calculation being used in sim_runner, but it does not seem to be used in [`request_observation`](https://github.com/lsst/rubin_scheduler/blob/55c8cf1d3a14d8da14658cc9459ebf42c3fce86f/rubin_scheduler/scheduler/schedulers/core_scheduler.py#L243). 
The observation that is sent here is the basis of "next visit" in the scheduler queue, so must have the actual expected rotSkyPos (more or less). If @tribeiro is overwriting that rotSkyPos/skyAngle downstream, that would be helpful to know, but it's not clear to me what information is contained in the "observation" that is requested (looks like it should have rotTelPos plus rotSkyPos, but these are not being converted via the rotation_converter presently).

rhiannonlynne:
The ZeroRotDetailer needs to be updated (both docstring and calculation of rotskyPos/rotTelPos in the class). Or dropped if it's unused, although a fixed-angle rotTelPos detailer seems like it could be useful. Same thing with the SpiderRotDetailer and Conccam90RotDetailer.

rhiannonlynne (commented):
Needs clarification on what rotSkyPos value is returned with `request_observation`.

yoachim:
The `CoreScheduler` is only doing an optional check on the rotation value to make sure an observation hasn't ended up beyond the rotator limits. Any rotSkyPos/rotTelPos values should have been set earlier by `Survey` or `Detailer` objects.

rhiannonlynne (approved):

PR #66, opened 2024-05-30
OPSIM-1154 scripted survey check filter loaded

yoachim:
A fix so scripted surveys do not propose observations where the needed filter is not loaded.


ehneilsen (approved):
Looks fine to me.

PR #69, opened 2024-06-24
 tickets/PREOPS-5256: make rubin_scheduler.scheduler.utils.SchemaConverter.opsim2obs robust to columns of unexpected type

ehneilsen:
The OR4 V4 opsim DB has a `psudoParaAngle` column of type `object` and value `None`, and `opsim2obs` throws an unhandled exception when it tries to convert this to degrees. Instead, catch the exception and set it to all `nan`.


rhiannonlynne:
So - before, we had the problem that columns didn't exist. 
And this handles the case where they exist but are improperly populated. Looks fine.

rhiannonlynne (approved):

PR #71, opened 2024-06-28
tickets/PREOPS-5190: Automatically run one-night auxtel sims daily to support the prenight briefing

ehneilsen:
This cron job needs to be update manually for new configurations. Automatic coordination of versions with the observatory version is PREOPS-5280.

### DIFF
--- a/batch/run_prenight_sims.sh
+++ b/batch/run_prenight_sims.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+#SBATCH --account=rubin:developers      # Account name
+#SBATCH --job-name=auxtel_prenight_daily   # Job name
+#SBATCH --output=/sdf/data/rubin/user/neilsen/batch/auxtel_prenight_daily/daily.out # Output file (stdout)
+#SBATCH --error=/sdf/data/rubin/user/neilsen/batch/auxtel_prenight_daily/daily.err  # Error file (stderr)
+#SBATCH --partition=milano              # Partition (queue) names
+#SBATCH --nodes=1                       # Number of nodes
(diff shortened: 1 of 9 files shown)


rhiannonlynne:
There's a problem with the unit tests, where `lsst.ts` is missing as an importable module. 

In general .. how committed are you to running and implementing things in this way?

ehneilsen:
> There's a problem with the unit tests, where `lsst.ts` is missing as an importable module.
> 
> In general .. how committed are you to running and implementing things in this way?

By "in this way", what in particular do you mean?

I think we definitely want to run a set of sims automatically every night, and the results need to be archived somewhere the reports can find them.

Aside from that, I'm attempting to do the simplest thing that will work. 

As far as the best way to get the right version goes (which I think is what I use lsst.ts for), I think the best think would be for the code needed to run the sims to be included in the container they are actually running at the summit, and have the cron job at the USDF grab that exact container and use it. At present, though, I don't know how to get access to it.

rhiannonlynne:
> As far as the best way to get the right version goes (which I think is what I use lsst.ts for), I think the best think would be for the code needed to run the sims to be included in the container they are actually running at the summit, and have the cron job at the USDF grab that exact container and use it. At present, though, I don't know how to get access to it.

"in this way" meant as in the PR, using GitPython to clone the repo and so on. 
I agree that getting the container (which is probably deployed via phalanx) or creating a new container built on that one which runs the simulations would likely be best. 
And also I agree that this is the simplest solution to get going for now.

rhiannonlynne:
The documentation build failed, but looks like the occasional problem with uploading documentation. This seems fine to merge.

rhiannonlynne (approved):

PR #74, opened 2024-07-17
Tickets/opsim 1164

yoachim:
Just getting rid of a few deprecation warnings and having `TestMakeSnapshot` actually skip if lsst.ts is not installed.

### DIFF
--- a/tests/sim_archive/test_make_snapshot.py
+++ b/tests/sim_archive/test_make_snapshot.py
@@ -2,9 +2,14 @@ import importlib.util
 import unittest
 
 from rubin_scheduler.scheduler.schedulers.core_scheduler import CoreScheduler
-from rubin_scheduler.sim_archive import get_scheduler_instance_from_repo
 
-HAVE_TS = importlib.util.find_spec("lsst.ts") is not None
+if importlib.util.find_spec("lsst"):
+    HAVE_TS = importlib.util.find_spec("lsst.ts")
+else:
+    HAVE_TS = False
+
+if HAVE_TS:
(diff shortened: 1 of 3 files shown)


tests/sim_archive/test_make_snapshot.py
```diff
@@ -2,9 +2,12 @@
 import unittest
 
 from rubin_scheduler.scheduler.schedulers.core_scheduler import CoreScheduler
-from rubin_scheduler.sim_archive import get_scheduler_instance_from_repo
 
-HAVE_TS = importlib.util.find_spec("lsst.ts") is not None
+try:
+    HAVE_TS = importlib.util.find_spec("lsst.ts") is not None
+    from rubin_scheduler.sim_archive import get_scheduler_instance_from_repo
```
ehneilsen, line 8: I've gotten style complaints (from ruff, I think) about using try/excepts to detect whether we have a module, with the suggestion to use find_spec instead.
So, I think what we need is something like
```
HAVE_TS = importlib.util.find_spec("lsst.ts") is not None
if HAVE_TS:
    from rubin_scheduler.sim_archive import get_scheduler_instance_from_repo
```
Yes, I think the try/except approach is more intuitive.

ehneilsen (approved):

tests/sim_archive/test_make_snapshot.py
yoachim: I think that only works if there is an `lsst` module. But that's easy enough to check first and still get rid of the try/except.

PR #75, opened 2024-07-18
add switch to use old ddf locs

yoachim:
Added missing return. Fixed mixed camel and snake case name.


rhiannonlynne (approved):
Much better name.

PR #76, opened 2024-07-18
OPSIM-1162:  comcam sv survey prep

rhiannonlynne:
Update docstrings on basis functions and features to clarify potential uses.
Also add new basis functions, detailers, and filter schedulers needed to simulate comcam SV surveys.


yoachim:
I added backwards compatibility so the v3.5 sims can run without modification. The unit tests are passing, but it looks like the v3.5 baseline (https://github.com/lsst-sims/sims_featureScheduler_runs3.5/blob/main/baseline/baseline.py) now fails with the `RuntimeError: Scheduler has failed to provide a valid observation multiple times...`, so I'm guessing there's a basis function that got changed in some way that's not quite right.

rhiannonlynne:
Ok, I'll have a look. 
Backwards compatibility - excellent idea.

yoachim:
Looks like the baseline simulation has similar but not identical results when run with this branch.

If that's going to happen I think we need to break this up into different PRs so we an keep track of when and where those changes are happening.

rhiannonlynne:
Yeah, not unreasonable. 
I'll pull out all of the name changes, the addition of the new fields, and the modification of survey_note/scheduler_note kwargs I guess (and any others that seem like they should make no change). 
And then have a separate PR just for the telalt/telaz part which is the basis function and the kinematic model. 
I'm not entirely certain where the comcam parts go in that, but I'll look.

rhiannonlynne:
The changes here were broken into separate PRs and incorporated through those. Closing this PR.

PR #82, opened 2024-08-09
convert observation schema from target to target_name

yoachim:
update "target" to "target_name" on schema for `observation` object to match downstream schema.


ehneilsen (commented):
Do we also need some code for backward compatibility, for example to load old output files with SchemaConverter?

ehneilsen (approved):
Look fine to me.

PR #87, opened 2024-08-27
OPSIM-1174: modify FieldSurvey, add new fields to observation

rhiannonlynne:
Add "science_program", "observation_reason" and "json_block" to observation
Modify FieldSurvey so it can accept only particular scheduler_note and depends directly on ScriptedSurvey.

### DIFF
--- a/rubin_scheduler/scheduler/surveys/field_survey.py
+++ b/rubin_scheduler/scheduler/surveys/field_survey.py
@@ -1,72 +1,316 @@
 __all__ = ("FieldSurvey",)
 
+import copy
+from functools import cached_property
+
 import numpy as np
 
 from rubin_scheduler.utils import ra_dec2_hpid
 
-from .dd_surveys import DeepDrillingSurvey
+from ..features import NObsSurvey
+from ..utils import empty_observation
+from . import BaseSurvey
 
 
-class FieldSurvey(DeepDrillingSurvey):
+class FieldSurvey(BaseSurvey):
+    """A survey class for running field surveys.
+
+    Parameters
+    ----------
+    basis_functions : `list` [`rubin_scheduler.scheduler.basis_function`]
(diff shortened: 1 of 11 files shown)


tests/scheduler/test_utils.py
```diff
@@ -59,7 +60,6 @@ def test_example(self):
             "pair_15, ri, b",
             "pair_15, yy, a",
             "pair_15, yy, b",
-            "pair_15, zy, a",
```
yoachim, line 62: Why not check for this one?

rubin_scheduler/scheduler/surveys/field_survey.py
```diff
@@ -1,72 +1,324 @@
...
+        target_name=None,
+        scheduler_note=None,
+        readtime=2.4,
         filter_change_time=120.0,
         nside=None,
         flush_pad=30.0,
-        seed=42,
         detailers=None,
     ):
+        default_nvisits = {"u": 20, "g": 20, "r": 20, "i": 20, "z": 20, "y": 20}
```
yoachim, line 88: Mutables as kwarg defaults is frowned upon, but I think it's more readable and like them.

yoachim (approved):

tests/scheduler/test_utils.py
rhiannonlynne: I wasn't seeing it come up. It may have been part of different changes .. I'll check what's being scheduled and put it back if it's present.

rubin_scheduler/scheduler/surveys/field_survey.py
rhiannonlynne: Are you asking me to change this, or leave it as is? the kwarg is nvisits, but it will fall back to default_nvisits which is an attribute if it's not set.

PR #89, opened 2024-08-28
PREOPS-5366: remove old minimum pair time kwarg since blob no longer adjust pair time

yoachim:
Fix issue where flush times were being set too short (because they were being computed in two places).


rhiannonlynne:
There are some other kwargs for the blob or other surveys that aren't used. While it doesn't have to go in this ticket, we should probably start doing that kind of thing more regularly. I think I have a commit somewhere that does a non-aggressive version of this, but that needs to folded in on some PR -- it's non-aggressive because I don't necessarily know what really can be dropped (such as the twilight scaling).

rhiannonlynne (approved):

PR #94, opened 2024-09-04
OPSIM-1181: backwards compatibility for field survey kwargs

rhiannonlynne:
Downstream failures in ts_fbs_utils, so replacing the kwarg for now.


tribeiro (approved):
Looks good.

PR #95, opened 2024-09-06
tickets/OPSIM-1184: use ra/dec to order

yoachim:
Eliminate an unnecessary ra,dec to alt,az conversion. Was probably originally stuck in there at some point because of a degrees,radians issue downstream.


ehneilsen (approved):
Looks fine, as far as I can tell. A test to verify that the change doesn't change the resultant order for some sample sets of observations would be reassuring.

PR #96, opened 2024-09-11
OPSIM-1188: remove comcam sv experiments, go to just simple greedy/pairs/field survey examples

rhiannonlynne:
Comcam on sky commissioning surveys will be laid out in ts_fbs_utils and ts_config_ocs, so having slightly different versions here could lead to confusion. 
I've removed these, but left a simple_field_survey utility that would replicate most of the same behavior for ease of testing various things. It also provides a template for building a field survey. 

~I've pulled out json_block from everywhere, as this was decided to mix too much of the scheduler_csc level information into rubin_scheduler.~
~I did leave target_name, science_program and observation_reason embedded in the field survey class as they are needed already .. they are not present in other surveys (so will be blank in the resulting observation, in general) .. this will be pending @yoachim 's work in https://rubinobs.atlassian.net/browse/OPSIM-1186~

With the merge of OPSIM-1186, I just tweaked the simple surveys in example_simple_survey to match and removed commits related to anything science_program or json_block.


rhiannonlynne:
I did have to fix a bug with FieldSurvey and its unit test related to "add_observation"  now that we're passing arrays and not elements of the array.

yoachim (approved):
Looks good. Since you need to rebase anyway might as well squash the boring black and rough commits.

PR #97, opened 2024-09-11
OPSIM-1182: Refactor add_observation to use an observation array (like empty_observation). 

yoachim:
Refactor so `add_observation` method takes a `observation` array, not just a slice. Added a check in `CoreScheduler` so it should be backwards compatible in case folks are passing slices.


rhiannonlynne:
Does the unit test `test_NObservationsCurrentSeason` need an update too? It looks like the unit test is using both add_observation(observation[0]) and update_season(observation[0]).

rhiannonlynne:
Guess we should add a deprecation warning to `CoaddedDepth` (feature) as it doesn't have an add_observations_array (which indicates we're not using it, generally).

rhiannonlynne:
Shoot - also, it seems like where I've made an observations_list and then iterated through the list, for testing features or anything below the core scheduler level, doing 
```
for obs in observations_list:
               <feature>/<basis_function>/<survey>.add_observation(obs)
```
is acting as if it were like the "observation[0]" instead of "observation" .. if that makes sense. |
This includes `test_NObsSurvey`, `test_LastObservation`, `test_NObservations`

yoachim:
> Shoot - also, it seems like where I've made an observations_list and then iterated through the list, for testing features or anything below the core scheduler level, doing
> 
> ```
> for obs in observations_list:
>                <feature>/<basis_function>/<survey>.add_observation(obs)
> ```
> 
> is acting as if it were like the "observation[0]" instead of "observation" .. if that makes sense. | This includes `test_NObsSurvey`, `test_LastObservation`, `test_NObservations`

These are all working fine and passing fully-formed `observation` arrays like they should.

rhiannonlynne:
Ah good - actually, now I see that those were ones you'd already changed too.

rhiannonlynne (approved):

PR #98, opened 2024-09-11
info into detailer

yoachim:
add `target_name`, `science_program`, and `observation_reason` as `Survey` kwargs. Any of them set on init will then go into a `Detailer` object that updates all observations coming out of the `Survey`.

### DIFF
--- a/rubin_scheduler/scheduler/surveys/surveys.py
+++ b/rubin_scheduler/scheduler/surveys/surveys.py
@@ -35,6 +35,7 @@ class GreedySurvey(BaseMarkovSurvey):
         camera="LSST",
         area_required=None,
         fields=None,
+        **kwargs,
     ):
         extra_features = {}
 
@@ -58,6 +59,7 @@ class GreedySurvey(BaseMarkovSurvey):
             camera=camera,
             area_required=area_required,
             fields=fields,
+            **kwargs,
         )
 
     def _generate_survey_name(self):
@@ -93,11 +95,6 @@ class GreedySurvey(BaseMarkovSurvey):
                 obs["nexp"] = self.nexp
                 obs["exptime"] = self.exptime
                 obs["scheduler_note"] = self.scheduler_note
-                obs["target_name"] = self.target_name
-                obs["science_program"] = self.science_program
-                obs["observation_reason"] = self.observation_reason
-                obs["json_block"] = self.json_block
-
                 observations.append(obs)
                 break
             iter += 1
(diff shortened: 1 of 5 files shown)


yoachim:
Looking at it now, we could toss `survey_name` in there as well. Then we just need to be sure folks know `survey_name` does not necessarily propagate directly to `scheduler_note`.

rubin_scheduler/scheduler/surveys/surveys.py
```diff
@@ -35,6 +35,7 @@ def __init__(
         camera="LSST",
         area_required=None,
         fields=None,
+        **kwargs,
```
rhiannonlynne, line 38: I guess since the BaseSurvey does not accept random kwargs, this is probably ok now, but this is how we ended up with random keyword arguments that did nothing yet were expected to do things.

rubin_scheduler/scheduler/surveys/surveys.py
```diff
@@ -196,6 +193,9 @@ def __init__(
         search_radius=None,
         alt_max=-9999,
         az_range=-9999,
+        target_name="",
```
rhiannonlynne, line 196: Should target_name be None or should it be ""? "" is almost None, but has different behavior if we're checking for it downstream.

rubin_scheduler/scheduler/surveys/surveys.py
```diff
@@ -196,6 +193,9 @@ def __init__(
...
+        observation_reason="FBS",
+        science_program="FBS",
```
rhiannonlynne, line 198: science_program is going to end up being the JSON BLOCK identifier, so it might be better if we don't set a default here .. although this is also probably something that is useful to discuss with Tiago. 
Leaving as "FBS" for now, when we're not actually using this, is probably also fine. Could replace both `science_program` and `observation_reason` with None for now as well.

rhiannonlynne:
I guess I'd prefer `survey_name` to not go into the detailer unless we think we gain some useful information? 
`scheduler_note` is clearly not `survey_name` for many of our baseline surveys.  (like the blob survey or the DD surveys, or the blob_long survey), and we probably shouldn't assume it will be for the FieldSurvey either. 
`survey_name` gets propagated to us being able to easily identify the Survey that it came from in any debugging .. visual example kind of thing here https://usdf-rsp-dev.slac.stanford.edu/schedview-snapshot/dashboard (pull up "Main" on 2024-09-10 15:11:21 and the first snapshot for an example .. here the survey_name is the BLOCK ID because that's how Tiago has to manage things at the moment, but I hope it will be more informative in the future).

rhiannonlynne (approved):

PR #99, opened 2024-09-11
get rid of confusing mjd_start in sim_runner

yoachim:
Get rid of confusing kwarg that didn't really work right anyway.


ehneilsen:
Can we use the change from u/neilsen/silly_start instead? The approach on u/yoachim/silly_start breaks stuff in sim archive, and following the approach in u/neilsen/silly_start makes it much easier to update it so it still works, and also avoids overloading the use of mjd_start.

yoachim:
Works for me.

PR #100, opened 2024-09-12
remove extras from Conditions object

yoachim:
Remove unused attributes from `Conditions` object.

### DIFF
--- a/rubin_scheduler/scheduler/features/conditions.py
+++ b/rubin_scheduler/scheduler/features/conditions.py
@@ -16,9 +16,7 @@ from rubin_scheduler.utils import (
     _approx_ra_dec2_alt_az,
     _hpid2_ra_dec,
     calc_lmst,
-    calc_season,
     m5_flat_sed,
-    survey_start_mjd,
 )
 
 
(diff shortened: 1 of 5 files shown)


rhiannonlynne:
Oh wow, I really thought something was using the seasons information in the conditions object -- however, I looked at all of the features which I thought were using it and no .. they hold it internally. Maybe it was at some intermediate point of development? Maybe some of the footprint basis functions did at some point?  Anyway, it's good to clear it out. I'm pretty sure nothing in ts_fbs_utils uses these, but if you could double-check that would be helpful. (I don't actually know if we would see any failures, unless you also try to run the schedulers it creates though).

rubin_scheduler/scheduler/features/conditions.py
```diff
@@ -572,7 +520,7 @@ def az_to_antisun(self):
         return self._az_to_antisun
 
     def __repr__(self):
-        return f"<{self.__class__.__name__} mjd_start='{self.mjd_start}' at {hex(id(self))}>"
+        return f"<{self.__class__.__name__} at {hex(id(self))}>"
```
rhiannonlynne, line 523: It might be useful to have this print the current time (not sure why it was mjd_start before, but conditions.mjd might be a good shortcut).

rhiannonlynne (approved):
I think this looks good.

PR #102, opened 2024-09-13
Tickets/opsim 1157: Change observations to proper numpy subclass

yoachim:
Make `ObservationArray` and `ScheduledObservationArray` subclasses of numpy arrays. Also dropping `JSON_Block` from the schema.

Quick test shows there's no performance hit making the change. Previous convienince functions updated to be backwards compatible and throw a warning.


rhiannonlynne (approved):
Looks good!

PR #105, opened 2024-09-17
Tickets/opsim 1191

yoachim:
Fix so `Conditions` object displays properly in a notebook and taking care of a warning that was triggering too aggressively.


ehneilsen (approved):
looks fine to me.

PR #106, opened 2024-09-17
OPSIM-1192: make_rolling_footprint work with default values

yoachim:
`make_rolling_footprint` now works with default kwarg values. Had to shuffle a few things around to avoid circular imports. Removed a strange duplicate `_is_in_ra_range` function (strange ruff never noticed that). Added bare-bones unit test for `make_rolling_footprint`.

### DIFF
--- a/rubin_scheduler/scheduler/utils/footprints.py
+++ b/rubin_scheduler/scheduler/utils/footprints.py
@@ -4,7 +4,6 @@ appropriately for a given time.
 """
 
 __all__ = (
-    "ra_dec_hp_map",
     "calc_norm_factor",
     "calc_norm_factor_array",
     "StepLine",
@@ -26,7 +25,7 @@ from astropy.coordinates import SkyCoord
 
 from rubin_scheduler.utils import _hpid2_ra_dec
 
-from .utils import set_default_nside
+from .sky_area import CurrentAreaMap
 
 
 def make_rolling_footprints(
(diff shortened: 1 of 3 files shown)


rhiannonlynne:
Could you please paste in some screenshots demonstrating that the rolling footprint is starting up in both uniform and not-uniform mode as expected? And also add screenshots of the footprint leading up to year 4 and year 7.

rubin_scheduler/scheduler/utils/utils.py
```diff
@@ -43,6 +44,16 @@
...
 
 
+def ra_dec_hp_map(nside=None):
+    """
+    Return all the RA,dec points for the centers of a healpix map, in radians.
+    """
+    if nside is None:
+        nside = set_default_nside()
+    ra, dec = _hpid2_ra_dec(nside, np.arange(hp.nside2npix(nside)))
+    return ra, dec
```
rhiannonlynne, line 54: Why is the original call not just replaced with _hpid2_radec?

ehneilsen (approved):
Looks fine to me.

ehneilsen:
There've been recent changes to `rubin_scheduler` that required changes to `schedview`. If you update your local `rubin_scheduler` and rebase your changes to `schedview`, it should work and you should be testing with the same versions github is.

yoachim:
> Could you please paste in some screenshots demonstrating that the rolling footprint is starting up in both uniform and not-uniform mode as expected? And also add screenshots of the footprint leading up to year 4 and year 7.

That's not what this PR is addressing.

PR #107, opened 2024-09-18
Bring documentation up to date

rhiannonlynne:
Update schema so it works again, and matches current schema. 
Add archive to the pages that display.
Add other examples in rubin_scheduler.scheduler.example into the API docs.


ehneilsen (approved):
Looks fine to me.

PR #108, opened 2024-09-18
tiickets/OPSIM-1193: add SURVEY_START_MJD and DEFAULT_NSIDE to replace functions

yoachim:
Introduce `SURVEY_START_MJD` and `DEFAULT_NSIDE` to replace `survey_start_mjd` and `set_default_nside`.


ehneilsen (approved):
Looks good, but there's one gotcha you should be aware of. I see you removed set_default_nside, presumably because one can now just assign DEFAULT_NSIDE to whatever they want. I think this will work just as well as using set_default_nside, *but* there are a number of places one would expect either to work when neither will. In particulary, if the call signature of a function explicitly references DEFAULT_NSIDE, e.g.
`def my_func(nside=DEFAULT_NSIDE):`
This default value will *not* get updated; DEFAULT_NSIDE will get refenced at the time of import, and never again. If you really want it to get appropriately updated, you need to do something along the lines of:
```
def my_func(nside=None):
    if nside is None:
        nside = DEFAULT_NSIDE
```
(The same applies to the start date, but there never was a way to change it anyway.)
I don't think the changes in this PR actually make anything worse than it was before, but it is an issue we should keep in mind.

yoachim:
I see, you're saying there's no way for the user to programmatically change `DEFAULT_NSIDE` and `SURVEY_START_MJD`. I think that's fine, if you want to run a different nside it's just like it's always been and you have to make sure to pass that new kwarg around to everything that needs it.

PR #109, opened 2024-09-23
let unscheduled downtimes be less than a day

yoachim:
Update `ModelObservatory` so it can handle downtime events that are shorter than 1 day.

### DIFF
--- a/tests/scheduler/test_modelobs.py
+++ b/tests/scheduler/test_modelobs.py
@@ -43,8 +43,8 @@ class TestModelObservatory(unittest.TestCase):
 
         mjd_down = mo_default.downtimes["start"][0] + 0.01
 
-        assert ~mo_default.check_up(mjd_down)
-        assert mo_ideal.check_up(mjd_down)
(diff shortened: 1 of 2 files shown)


tests/scheduler/test_modelobs.py
```diff
@@ -43,8 +43,8 @@ def test_ideal(self):
 
         mjd_down = mo_default.downtimes["start"][0] + 0.01
 
-        assert ~mo_default.check_up(mjd_down)
-        assert mo_ideal.check_up(mjd_down)
+        assert ~mo_default.check_up(mjd_down)[0]
+        assert mo_ideal.check_up(mjd_down)[0]
```
rhiannonlynne, line 47: It would be great to add a test with a short downtime and see that the time returned by check_mjd is the appropriate time, and that if you land in the middle of a day, that it advances to the correct time of sunset.

rhiannonlynne (commented):
The update looks correct, but it would be good to expand the unit test to confirm that it stays correct.

rhiannonlynne (approved):

PR #112, opened 2024-10-03
remove core sched rotator checking. add whole queue kwarg to request_…

yoachim:
Remove some rotator checking that's in `CoreScheduler`.
Add a kwarg to return the whole observing queue rather than a single observation.

### DIFF
--- a/rubin_scheduler/scheduler/schedulers/core_scheduler.py
+++ b/rubin_scheduler/scheduler/schedulers/core_scheduler.py
@@ -12,13 +12,7 @@ import pandas as pd
 from astropy.time import Time
 
 from rubin_scheduler.scheduler.utils import HpInComcamFov, HpInLsstFov, IntRounded, ObservationArray
-from rubin_scheduler.utils import (
-    DEFAULT_NSIDE,
-    _approx_altaz2pa,
-    _approx_ra_dec2_alt_az,
-    _hpid2_ra_dec,
-    rotation_converter,
-)
+from rubin_scheduler.utils import DEFAULT_NSIDE, _hpid2_ra_dec, rotation_converter
 
 
 class CoreScheduler:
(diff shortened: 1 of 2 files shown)


rubin_scheduler/scheduler/schedulers/core_scheduler.py
```diff
@@ -213,21 +203,25 @@ def _check_queue_mjd_only(self, mjd):
...
             The Modified Julian Date.
             If None, it uses the MJD from the conditions from the
             last conditions update.
+        whole_queue : `bool`
+            Return the entire list of observations in the queue (True),
+            or just a single observation (False). Default False
 
         Returns
         -------
         observation :  observation object (ra,dec,filter,rotangle)
```
rhiannonlynne, line 222: Is this returning an ObservationArray now?

rhiannonlynne: (or list of ObservationArray)?

rhiannonlynne:
I'm guessing checking the rotator angle wasn't needed because the surveys didn't propose out of range rotation angles? Or is it just that the checking shouldn't happen here, and instead in the ModelObservatory/SchedulerCSC? 
What happens if a survey doesn't have a detailer to set rotation angles? (some of the auxtel surveys send '[]' and not None, which doesn't trigger the addition of the ZeroRotation detailer, but also I don't think that's quite what they want for the spectroscopic surveys ... but also i think those are probably having rotation angles added after the fact in some other CSC).

rhiannonlynne:
This probably wants @tribeiro to comment.

rubin_scheduler/scheduler/schedulers/core_scheduler.py
yoachim: List of ObservationArray.

rhiannonlynne: Can you update this docstring please? 
probably something like 
observation : `list` [`~.scheduler.utils.ObservationArray`]

rhiannonlynne:
@tribeiro said on slack that this is good for him - 
so looks good.

rhiannonlynne (approved):

PR #113, opened 2024-10-03
Tickets/opsim 1197

yoachim:
Updates to `Conditions` object. Moved airmass calc into `Conditions` (so it doesn't need to be manually set). Add convenience functions for updating attributes en mass.

### DIFF
--- a/rubin_scheduler/scheduler/features/conditions.py
+++ b/rubin_scheduler/scheduler/features/conditions.py
@@ -158,8 +158,6 @@ class Conditions:
             during the current night. From interpolation.
         sunrise : `float`
             The MJD of sunrise during the current night. From interpolation
-        mjd_start : `float`
-            The starting MJD of the survey.
         moonrise : `float`
             The MJD of moonrise during the current night. From interpolation.
         moonset : `float`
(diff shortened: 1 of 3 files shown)


rubin_scheduler/scheduler/features/conditions.py
```diff
@@ -380,12 +379,20 @@ def slewtime(self, value):
 
     @property
     def airmass(self):
+        if self._airmass is None:
```
ehneilsen, line 382: Why not use `@cached_property` instead of the funny dance with `self._airmass`?
If `self.airmass` is your decorated method that computes airmass, and you need to clear the cache, you can delete the property: `del self.airmass`.
The next time you reference `self.airmass` it will get recomputed (and re-cached with the new value).
If you just want to assign it to something that isn't computed by your decorated method, you can just set the cache just by setting the property: 
`self.airmass = 1.5` will set the cache to 1.5.

ehneilsen (approved):
Only minor suggestions.

rubin_scheduler/scheduler/features/conditions.py
```diff
@@ -516,6 +523,98 @@ def az_to_antisun(self):
...
+        for key in auxtel_args.parameters.keys():
+            setattr(self, key, loc[key])
+
+        potential_attrs = dir(self)
+        for key in kwargs:
+            if key not in potential_attrs:
+                warnings.warn("Setting unexpected Conditions attribute %s" % key)
+            setattr(self, key, kwargs[key])
+
+    def set_maintel_info(
```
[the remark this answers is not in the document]
yoachim: Naming ceremony isn't until tomorrow, I'm sneaking it in under the wire!

rubin_scheduler/scheduler/features/conditions.py
yoachim: Mostly it makes it analogous with all the other attributes. This way if someone changes the MJD the airmass gets reset to `None` like everything else.

PR #114, opened 2024-10-07
Tickets/sp 1619

yoachim:
New HEALpix utility function for growing a mask. Should be much faster than healpy smoothing method.

### DIFF
--- a/tests/utils/test_healutils.py
+++ b/tests/utils/test_healutils.py
@@ -77,6 +77,31 @@ class TestHealUtils(unittest.TestCase):
         self.assertEqual(map3[hpid], 0.0)
         self.assertEqual(hp.maptype(map3), 0)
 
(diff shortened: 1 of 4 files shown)


tests/utils/test_healutils.py
ehneilsen, line 1: It would be reassuring if there were a test here that directly compared the results with those that would be produced by the old hp.smoothing method.

ehneilsen (approved):
Looks fine to me, and does indeed speed up make_reward_df.

tests/utils/test_healutils.py
yoachim: I'm not too worried about it being an exact match to the old method, but putting in a unit test that does an explicit check that sizes are correct is a good idea. And that caught a potential error with the caching.

PR #115, opened 2024-10-08
fix healpy mask growth to work beyond small angle approx

yoachim:
Mask growth only worked properly for small angles, now works for any mask padding size. Extended unit tests to include angles on many scales.


rubin_scheduler/utils/healpy_utils.py
```diff
@@ -398,14 +398,20 @@ def _hp_grow_mask(nside, masked_indx_tuple, grow_size=np.radians(2.0), scale=100
     # get the points that way and save an _hpid2_ra_dec call.
     ra, dec = _hpid2_ra_dec(nside, np.arange(hp.nside2npix(nside))[list(masked_indx_tuple)])
     x, y, z = _xyz_from_ra_dec(ra, dec)
+
+    # convert angular distance to Euclidian distance
```
ehneilsen, line 402: Wouldn't this just be dist = 2*sin(grow_size/2) ?

ehneilsen (approved):
looks fine

rubin_scheduler/utils/healpy_utils.py
yoachim: Yes. Yes it is.

PR #116, opened 2024-10-09
Tickets/sp 1629

yoachim:
Updating example scheduler to match the v4.0 survey simulation. 

Adds utilities for generating ToO events of various types.

### DIFF
--- a/rubin_scheduler/scheduler/example/example_scheduler.py
+++ b/rubin_scheduler/scheduler/example/example_scheduler.py
@@ -20,8 +20,6 @@ import healpy as hp
 import numpy as np
 from astropy import units as u
 from astropy.coordinates import SkyCoord
-
-# So things don't fail on hyak
 from astropy.utils import iers
 
 import rubin_scheduler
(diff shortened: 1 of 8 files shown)


rubin_scheduler/scheduler/example/example_scheduler.py
```diff
@@ -517,25 +523,26 @@ def gen_long_gaps_survey(
             u_template_weight=u_template_weight,
             g_template_weight=g_template_weight,
             blob_names=blob_names,
+            u_exptime=u_exptime,
+            nexp=nexp,
         )
         scripted = ScriptedSurvey(
-            [bf.AvoidDirectWind(nside=nside), bf.AltAzShadowMaskBasisFunction(nside=nside, shadow_minutes=0)],
+            [bf.AvoidDirectWind(nside=nside)],
```
rhiannonlynne, line 530: How does the ScriptedSurvey for the followup triplets know what parts of the sky are unavailable for pointing due to the telescope?

rubin_scheduler/scheduler/example/example_scheduler.py
```diff
@@ -1133,20 +1141,20 @@ def ddf_surveys(detailers=None, season_unobs_frac=0.2, euclid_detailers=None, ns
...
 
 
-def ecliptic_target(nside=SURVEY_START_MJD, dist_to_eclip=40.0, dec_max=30.0, mask=None):
+def ecliptic_target(nside=DEFAULT_NSIDE, dist_to_eclip=40.0, dec_max=30.0, mask=None):
     """Generate a target_map for the area around the ecliptic
 
     Parameters
     ----------
-    nside : int (DEFAULT_NSIDE)
+    nside : int
```
ehneilsen, line 1149: At some point some formatting work is needed in rubin_scheduler to fix the markup so the auto-generated docs have working links, for example by putting backticks around the types for parameters.
I'm not convinced that in this PR is the right place to do this, though, but we should do this eventually.

ehneilsen (approved):
Okay, passes tests in schedview now too (if I update the schedview test data).

rubin_scheduler/scheduler/example/example_scheduler.py
yoachim: There's a `_check_alts_ha` method on `ScriptedSurvey`. It doesn't make as much sense to use HEALpix maps when there's already a list of RA,dec positions.

PR #117, opened 2024-10-10
tickets/SP-1623: refactor metadata storage in the prenight scheduler simulation archive to improve performance

ehneilsen:
This modification creates a cache of metadata in an hdf5 file in the same bucket as the sims.
When querying for metadata, the code now checks the cached metadata, and looks in the cache for metadata for any dates before the last one in the cache. For dates of the last one in the cache or later, it falls back to the older behavior of lookinng in the yaml metadata files for each sim. ("Of the last date" rather than "after the last date" in case the cache does not include all sims from the last date.)
A separate cron job updates the cache daily.
A corresponding branch of schedview_notebooks causes the prenight_sims notebook to use the cache.


rhiannonlynne:
Hi Eric, sorry - think I looked at this a while back and meant to come back to poke the docs build failure but forgot. 
The content of the PR seems fine .. I am a little confused whether "compilation of sim archive" is something that needs to be done once, and then not again, or not. 
Can you clarify this part?

rhiannonlynne (approved):

PR #118, opened 2024-10-17
docstring updating

yoachim:
Lots of docstring formatting in example_scheduler.py

### DIFF
--- a/rubin_scheduler/scheduler/example/example_scheduler.py
+++ b/rubin_scheduler/scheduler/example/example_scheduler.py
@@ -104,46 +104,52 @@ def standard_bf(
 
     Parameters
     ----------
-    nside : int (DEFAULT_NSIDE)
-        The HEALpix nside to use
-    nexp : int (1)
-        The number of exposures to use in a visit.
-    exptime : float (30.)
-        The exposure time to use per visit (seconds)
-    filtername : list of str
-        The filternames for the first set
-    filtername2 : list of str
-        The filter names for the second in the pair (None if unpaired)
-    n_obs_template : dict (None)
-        The number of observations to take every season in each filter
-    season : float (300)
-        The length of season (i.e., how long before templates expire) (days)
-    season_start_hour : float (-4.)
-        For weighting how strongly a template image needs to be
-        observed (hours)
-    sesason_end_hour : float (2.)
-        For weighting how strongly a template image needs to be
-        observed (hours)
-    moon_distance : float (30.)
-        The mask radius to apply around the moon (degrees)
-    m5_weight : float (3.)
-        The weight for the 5-sigma depth difference basis function
-    footprint_weight : float (0.3)
+    nside : `int`
+        The HEALpix nside to use. Defaults to DEFAULT_NSIDE
+    filtername : `str`
+        The filter name for the first observation. Default "g".
+    filtername2 : `str`
+        The filter name for the second in the pair (None if unpaired).
+        Default "i".
+    n_obs_template : `dict`
+        The number of observations to take every season in each filter.
+        Default None.
+    season : `float`
+        The length of season (i.e., how long before templates expire) (days).
+        Default 300.
+    season_start_hour : `float`
+        Hour angle limits to use when gathering templates.
+        Default -4 (hours)
+    sesason_end_hour : `float`
+       Hour angle limits to use when gathering templates.
+       Default +2 (hours)
+    moon_distance : `float`
+        The mask radius to apply around the moon (degrees).
+        Default 30.
+    m5_weight : `float`
+        The weight for the 5-sigma depth difference basis function.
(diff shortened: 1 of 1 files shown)


rubin_scheduler/scheduler/example/example_scheduler.py
```diff
@@ -326,56 +332,71 @@ def blob_for_long(
...
-        The ideal time between pairs (minutes)
-    camera_rot_limits : list of float ([-80., 80.])
+    nside : `int`
+        The HEALpix nside to use. Default to DEFAULT_NSIDE.
+    nexp : `int`
+        The number of exposures to use in a visit. Default 2.
+    exptime : `float`
+        The exposure time to use per visit (seconds).
+        Default 29.2
+    filter1s : `list` of `str`
```
rhiannonlynne, line 342: `list` [`str`]  rather than `list` of `str` (https://developer.lsst.io/python/numpydoc.html#py-docstring-parameter-types-sequences)

rubin_scheduler/scheduler/example/example_scheduler.py
```diff
@@ -498,8 +519,38 @@ def gen_long_gaps_survey(
     """
     Paramterers
     -----------
-    HA_min(_max) : float
+    footprints : `rubin_scheduler.scheduler.utils.footprints.Footprints`
+        The footprints to be used.
+    night_pattern : `list` of `bool`
```
rhiannonlynne, line 524: `list` [`bool`]

rubin_scheduler/scheduler/example/example_scheduler.py
```diff
@@ -565,32 +616,43 @@ def gen_greedy_surveys(
...
-    slewtime_weight : float (3.)
-        The weight on the slewtime basis function
-    stayfilter_weight : float (3.)
+        Default 0.3 (uniteless).
+    slewtime_weight : `float`
+        The weight on the slewtime basis function.
+        Default 3.0 (uniteless).
+    stayfilter_weight : `float`
         The weight on basis function that tries to stay avoid filter changes.
+        Default 3.0 (uniteless).
```
rhiannonlynne, line 655: unitless

rubin_scheduler/scheduler/example/example_scheduler.py
```diff
@@ -702,59 +764,76 @@ def generate_blobs(
...
-    exptime : float (30.)
-        The exposure time to use per visit (seconds)
-    filter1s : list of str
-        The filternames for the first set
-    filter2s : list of str
+        Default 1.
+    exptime : `float`
+        The exposure time to use per visit (seconds).
+        Default 29.2
+    filter1s : `list` of `str`
```
rhiannonlynne, line 775: `list` [`str`]

rubin_scheduler/scheduler/example/example_scheduler.py
```diff
@@ -104,46 +104,52 @@ def standard_bf(
...
-        observed (hours)
-    moon_distance : float (30.)
-        The mask radius to apply around the moon (degrees)
-    m5_weight : float (3.)
-        The weight for the 5-sigma depth difference basis function
-    footprint_weight : float (0.3)
+    nside : `int`
+        The HEALpix nside to use. Defaults to DEFAULT_NSIDE
+    filtername : `str`
+        The filter name for the first observation. Default "g".
```
rhiannonlynne, line 110: Where we have a default value, it will show up in the docstring when this is built in the docs (see first screenshot). Because the docstring and the call are right next to each other, we don't have to repeat the default values in the docstring (and, as you can see in the second screenshot, it can make it all look messier). 

I think if we're using None so that it defaults to some other value within the code, that "Default None will be set to xxxx" is useful, or  it can be useful to know if we're setting the default for a good reason that other people should know (and then probably not change) -  but otherwise we should probably get out of the habit of repeating our defaults. 

Screenshot 2024-10-21 at 7 06 12 PM
 
Screenshot 2024-10-21 at 7 08 24 PM

rhiannonlynne (approved):
`list` of `float` (etc) needs to change to `list` [`float`] (etc)

rubin_scheduler/scheduler/example/example_scheduler.py
yoachim: But if you pull up the docstring from the command line or in a notebook the call signature and the relevant section of the docstring can get totally separated. 

image

rhiannonlynne: I guess I was thinking that 
Screenshot 2024-10-22 at 2 21 22 PM
is still less readable than this (where removing the additional Default lines means that the whole thing is shorter)
Screenshot 2024-10-22 at 2 24 24 PM

My main problem is that the defaults that are in these sections tend to get outdated. If these are kept up to date, I'll be ok.

PR #120, opened 2024-10-25
tickets/SP-1667: support reading multiple nights

ehneilsen:
Previously, the consdb tools in rubin_scheduler could only query visits for one night at a time. To support DDF cadence plots (needed to merge the different version of the night summary for SP-1667), all nights to be included in the DDF plot should be queryable at once.

### DIFF
--- a/rubin_scheduler/utils/consdb.py
+++ b/rubin_scheduler/utils/consdb.py
@@ -110,10 +110,13 @@ class ConsDBVisits(ABC):
         The date for which to query the database.
     url : `str`
         The connection string from the database.
+    num_nights : `int`
+        The number of nights (ending in day_obs) for which to get visits.
     """
 
     day_obs: str | int
     url: str = "postgresql://usdf@usdf-summitdb.slac.stanford.edu:5432/exposurelog"
+    num_nights: int = 1
 
     @property
     @abstractmethod
@@ -314,7 +317,8 @@ class ConsDBVisits(ABC):
                 AND e.s_dec IS NOT NULL
                 AND e.sky_rotation IS NOT NULL
                 AND ((e.band IS NOT NULL) OR (e.physical_filter IS NOT NULL))
-                AND e.day_obs = {self.day_obs_int}
+                AND e.day_obs <= {self.day_obs_int}
+                AND e.day_obs > {self.day_obs_int - self.num_nights}
         """
         return query_consdb(consdb_visits_query, self.url)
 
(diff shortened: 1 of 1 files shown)


rhiannonlynne:
Ah - unfortunately, dayobsint doesn't work as a simple subtraction.  (i.e. 20240301 - 1 goes to 20240228 most of the time)
Merlin has all of this coded into summit_utils, but unfortunately (for us) this isn't a conda package. 
I have been doing this conversion by using Time + TimeDelta objects, then turning those back into dayobsint style values.

ehneilsen:
Ugh. I absolutely despise representing dates as a YYYYMMDD integer. As part of the SP-1667 branch in schedview I've added some tools to help deal with it. I'll do something here as well.

rhiannonlynne:
Internally, we could always just use Time objects and only convert when going to places where it makes sense.

rubin_scheduler/utils/consdb.py
```diff
@@ -305,6 +309,10 @@ def consdb_visits(self) -> pd.DataFrame:
         # and visit_id = exposure_id.
         # To support 2 exposures snaps in the future, this query will
         # need to be rewritten to group by visit_id.
+        day_obs_date = datetime.strptime(f"{self.day_obs_int}", "%Y%m%d").date()
+        prior_day_obs_date = day_obs_date - timedelta(days=self.num_nights)
+        prior_day_obs = int(prior_day_obs_date.strftime("%Y%m%d"))
```
rhiannonlynne, line 314: I find it much easier to keep track of what format is what for dayobs or dayobsint (or other numbers that are into or out of that format) if I always use the same "kind" of names. 
As in -- if prior_day_obs is the same kind of thing as day_obs_int, couldn't it be "prior_day_obs_int"? 
Otherwise though, this seems fine.

rubin_scheduler/utils/consdb.py
```diff
@@ -110,10 +111,13 @@ class ConsDBVisits(ABC):
         The date for which to query the database.
     url : `str`
         The connection string from the database.
+    num_nights : `int`
+        The number of nights (ending in day_obs) for which to get visits.
```
rhiannonlynne, line 115: ending in day_obs or "including day_obs" ? 
(just thinking 1 = day_obs itself, not the night before ..)

rhiannonlynne (approved):
I have some minor comments - for future maintainability, using "prior_day_obs_int" seems like it would be more consistent with the naming of "day_obs_int" and be easier to follow later. 
And ("ending in day_obs") in the docstring .. well, numbers of nights and so on is always confusing to word, but I think "including day_obs" is probably more obvious.

PR #121, opened 2024-10-28
tickets/SP-1685

ehneilsen:
Changes here are being used by the new schedview night summary Times Square notebook.

### DIFF
--- a/rubin_scheduler/utils/consdb.py
+++ b/rubin_scheduler/utils/consdb.py
@@ -308,10 +308,6 @@ class ConsDBVisits(ABC):
         consdb_visits : `pd.DataFrame`
             The visit data.
         """
-        # In the schema as of now, all visits are one exposure,
-        # and visit_id = exposure_id.
-        # To support 2 exposures snaps in the future, this query will
-        # need to be rewritten to group by visit_id.
         day_obs_date = datetime.strptime(f"{self.day_obs_int}", "%Y%m%d").date()
         prior_day_obs_date = day_obs_date - timedelta(days=self.num_nights)
         prior_day_obs_int = int(prior_day_obs_date.strftime("%Y%m%d"))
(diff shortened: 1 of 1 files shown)


rubin_scheduler/utils/consdb.py
```diff
@@ -316,8 +316,32 @@ def consdb_visits(self) -> pd.DataFrame:
...
+        ql_columns = (
+            query_consdb(columns_query(f"visit{self.num_exposures}_quicklook", self.instrument))
+            .loc[:, "column_name"]
+            .values
+        )
+        visit_query_exposure_columns = ", ".join([f"e.{c} AS {c}" for c in exposure_columns])
+        visit_query_ql_columns = ", ".join(
+            [f"v{self.num_exposures}q.{c} AS {c}" for c in ql_columns if c not in exposure_columns]
+        )
+        visit_query_columns = ", ".join([visit_query_exposure_columns, visit_query_ql_columns])
```
rhiannonlynne, line 340: We should standardize on using "visit1" table OR exposure table, but once choosing one, should name the subsequent use of those values match. 
Really, visit1 goes with visit1_quicklook, and matches the term "visits" so I'd prefer going with the visit1 table instead of exposure.
Exposure really is exposures .. if /when we do multiple snaps per visit, exposure will deviate significantly from visit1.

rubin_scheduler/utils/consdb.py
```diff
@@ -316,8 +316,32 @@ def consdb_visits(self) -> pd.DataFrame:
         prior_day_obs_date = day_obs_date - timedelta(days=self.num_nights)
         prior_day_obs_int = int(prior_day_obs_date.strftime("%Y%m%d"))
 
+        # To avoid duplicate columns, explicitly include list of columns to
+        # to return in the query instead of using a "SELECT *".
+        def columns_query(table, instrument):
+            return f"""
+                SELECT column_name FROM information_schema.columns
```
rhiannonlynne, line 323: Based on KT's reaction, might want to think about whether information_schema.columns is the way to go or whether it should be the ConsDB client schema(schema, table) call.  (it would however be interesting to check that you get the same values .. I see that some keys are missing from the values returned by that call). 
We may want to move into more consDB client usage as well, instead of raw postgres or sqlachemy .. the consdb client is intended to have some flex with changing columns I think (through some of its "flexible metadata" use), so may isolate a bit from changes in the column names due to changes in the butler. 
I've been ok using the ConsDB client directly.

rubin_scheduler/utils/consdb.py
```diff
@@ -316,8 +316,32 @@ def consdb_visits(self) -> pd.DataFrame:
...
+        )
+        visit_query_exposure_columns = ", ".join([f"e.{c} AS {c}" for c in exposure_columns])
+        visit_query_ql_columns = ", ".join(
+            [f"v{self.num_exposures}q.{c} AS {c}" for c in ql_columns if c not in exposure_columns]
+        )
+        visit_query_columns = ", ".join([visit_query_exposure_columns, visit_query_ql_columns])
+
         consdb_visits_query: str = f"""
-            SELECT * FROM cdb_{self.instrument}.exposure AS e
+            SELECT {visit_query_columns}
```
rhiannonlynne, line 343: These aren't visit columns if you're querying the exposure table, they're exposure columns.

rubin_scheduler/utils/consdb.py
```diff
@@ -956,6 +989,34 @@ def instrumental_zeropoint_for_band(self, band: str) -> float:
...
+        else:
+            warn(
+                "Cannot use either visit_id, exposure_id, or guess observation_id, just making something up!"
+            )
+            observation_id = pd.Series(
+                Time(self.consdb_visits.obs_start).strftime("9%Y%m%d%H%M%S").astype(int),
+                index=self.consdb_visits.index,
+            )
+
+        return observation_id
```
rhiannonlynne, line 1018: I'm curious - why is this needed? Have some queries from the consDB returned information with non-unique identifiers? 
Although, querying from the exposure table would probably not give unique visit_id values, or at least, it wouldn't be guaranteed to be unique (as there could be two exposures for a given visit).

rhiannonlynne (changes requested):
Unless we really want to look at separate exposures (and not "visits"), I think we should shift to the visit1 table instead of exposure table. It might make some of the rest a little simpler too. Are there things we need to run/test/monitor on an exposure level rather than visit level?

rubin_scheduler/utils/consdb.py
ehneilsen: I used information_schema.columns directly because that's what I already knew how to do; for our use, using the ConsDB client seems like learning and extra new API that doesn't really add anything I can't already do just as easily with raw postgresql.
I see some trade-offs in stability, and it's not clear which "wins" here. As you point out, maybe there will be fiddles in the ConsDB client that insulate code from schema changes in the database. On the other hand, querying  information_schema.columns is guaranteed to be correct (because that's how postgresql itself knows what columns are in the table), while it's possible for the EFDClient to be wrong about its schema (I think this has already happened), and using EFDClient also means that the code is vulnerable to changes in the EFDClient module itself, while direct postgresql queries are not.

rubin_scheduler/utils/consdb.py
ehneilsen: That's because some of the columns we need is present in the exposure table and not the visit table. I was when starting this for OR4 that joining the visit_id to the exposure_id would get the first exposure, which was good enough for what I needed (particularly given that for OR4 there was only one exposure per visit).

For arbitrary numbers of exposures per visit, the best thing is probably to find some kind of a join that would match all exposure corresponding to a visit to that visit, GROUP BY visit_id, and use SQL aggregation functions (probably just using the first exposure in each group). For OR4, such a join wasn't possible.

Double checking, it looks ilke it still isn't possible to do this join correctly, so multiple exposures match each visit. But, there still isn't a visit2 table, so I don't think multi-snap visits are supported by the consdb yet anyway.

However, it looks like most of the columns of interest are present in the visits1 table now anyway, so maybe the join to exposure is no longer necessary. I'll give that a try.

rhiannonlynne: If there are things which don't work about the consdb as it is now, I think the answer is probably to try to get it changed -- or at the very least, record and pass along what isn't working.

rubin_scheduler/utils/consdb.py
rhiannonlynne: Hm. Ok.

rubin_scheduler/utils/consdb.py
ehneilsen: Okay, yeah, I'm switching everything to visit1. It's visits we really care about, because the scheduler doesn't schedule exposures independently of visits.

rubin_scheduler/utils/consdb.py
ehneilsen: Whatever the issues was with the visits table in OR4 that drove me to use exposure instead, it seems to be solved now: I switched from using `exposures` to` f"visits{self.num_exposures}"`, and it seems to be working.

rubin_scheduler/utils/consdb.py
ehneilsen: Yes, I had an issue with missing ids that this addressed. 
Because of the way I did the join in the query, there could never be more that one exposure_id matched to a visit, because the join was on exposure_id = visit_id.
With the change I just made, though, exposure_id shouldn't be there any more, so I've removed that branch of the if sequence.

rhiannonlynne (approved):

PR #122, opened 2024-10-30
add target_name to observations competed by ToO surveys

yoachim:
Update ToO surveys so they populate the `target_name` field.


rhiannonlynne (approved):
Sweet - so this  puts "target_name" == the type of ToO + its number, right? 
This will be helpful for grouping observations of the same ToO together too.

PR #124, opened 2024-11-04
Make sure ToO surveys won't saturate if nsnaps is set to 1

yoachim:
Adding some additional kwargs for ToO setup to ensure we don't run into the regime where long exposures would saturate on the sky background.

### DIFF
--- a/rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
+++ b/rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
@@ -60,7 +60,7 @@ class ToOScriptedSurvey(ScriptedSurvey, BaseMarkovSurvey):
         Should long exposure times be split into multiple snaps.
         Default False.
     split_long_max : `float`
-        Maximum exposure time to allow before splitting into 
+        Maximum exposure time to allow before splitting into
         multiple snaps if split_long is True. Default 30s.
     split_long_div : `float`
         Time to divide the exposure time by to decide how many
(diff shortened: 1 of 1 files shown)


rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
```diff
@@ -358,8 +370,8 @@ def _new_event(self, target_o_o, conditions):
                             # check if we should break
                             # long exposures into multiple
                             if self.split_long:
-                                if exptime > 119:
-                                    nexp = int(np.round(exptime / 30.0))
+                                if exptime > self.split_long_max:
+                                    nexp = int(np.round(exptime / self.split_long_div))
```
rhiannonlynne, line 374: I don't understand how this is supposed to work. 
With the defaults, if there is a ToO with a 45 second exposure time, it will be rounded down to a 0s exposure won't it? 
Should split_long_max and split_long_div be the same? 
I also don't see how this avoids saturation in particular, given that the skybackground isn't being checked.

rhiannonlynne (changes requested):

rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
yoachim: Ah, that should be a `ceil` instead of a `round`.
We could make a full blown detailer to do sky saturation checking. Right now if we set the number of snaps to 1, the longer ToO observations saturate, so this is just making it so ToO number of snaps is no longer strictly tied to the general survey number of snaps.

rhiannonlynne (approved):

PR #125, opened 2024-11-13
SP-1722: add FieldAltAzSurvey

yoachim:
Add a field survey that uses alt,az rather then ra,dec.

### DIFF
--- a/rubin_scheduler/scheduler/surveys/field_survey.py
+++ b/rubin_scheduler/scheduler/surveys/field_survey.py
@@ -1,12 +1,13 @@
-__all__ = ("FieldSurvey",)
+__all__ = ("FieldSurvey", "FieldAltAzSurvey")
 
 import copy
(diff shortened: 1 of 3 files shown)


rhiannonlynne (commented):
Comments left in line

rubin_scheduler/scheduler/surveys/field_survey.py
```diff
@@ -198,9 +199,9 @@ def _generate_survey_name(self, target_name=None):
         else:
             self.survey_name = f"Field {self.ra_deg :.2f} {self.dec_deg :.2f}"
 
-    @cached_property
+    @property
```
rhiannonlynne, line 202: For *regular* field surveys, where the RA/Dec of the healpix desired doesn't change, this should be a cached_property. 
For *altaz* field surveys, it should *not* be cached, and just a @property.

PR #126, opened 2024-11-14
tickets/SP-1723: impove support for data from consdb

ehneilsen:


yoachim (approved):
looks good

rubin_scheduler/utils/consdb.py
```diff
@@ -520,7 +520,41 @@ def note(self) -> pd.Series:
         note : `pd.Series`
             Note for each visit.
         """
-        return self.consdb_visits["observation_reason"]
+        if "note" in self.consdb_visits:
+            note = self.consdb_visits["note"]
+        else:
+            note = pd.Series("", index=self.consdb_visits.index)
+        return note
```
rhiannonlynne, line 527: If "note" is really "scheduler_note" we should probably start using "scheduler_note" here as well as other places to avoid confusion.

rubin_scheduler/utils/consdb.py
```diff
@@ -867,59 +890,58 @@ def opsim(self) -> pd.DataFrame:
...
+            "azimuth": self.azimuth,
+            "paraAngle": self.parallactic_angle,
+            "psudoParaAngle": self.pseudo_parallactic_angle,
+            "cloud": self.cloud,
+            "moonAlt": self.sun_moon_positions["moon_alt"],
+            "sunAlt": self.sun_moon_positions["sun_alt"],
+            "note": self.note,
+            "scheduler_note": "NOT AVAILABLE",
+            "target_name": self.target_name,
+            #            "block_id": "NOT AVAILABLE",
```
rhiannonlynne, line 924: We may want to pass along block_id, as it's a valid thing in the consdb .. but we do need to think about what it means I suppose, given we have a field called this in the opsim outputs which is not the same thing as it is in the consdb outputs.

rhiannonlynne (approved):

rubin_scheduler/utils/consdb.py
ehneilsen: I went ahead and added `scheduler_note` as its own thing in addition to `note`, taking its value from `scheduler_note` from `consdb` (if it exists). Once `note` is removed from the [officially documented schema](https://rubin-scheduler.lsst.io/fbs-output-schema.html), we should also remove it here.

PR #127, opened 2024-11-17
Fix bug in CameraRotSmallDitherDetailer

rhiannonlynne:
CameraRotSmallDitherDetailer was losing the large offsets to rotation angle between filter changes after the first visit in each filter in a sequence, and going back to only keeping the small per-visit offset.

### DIFF
--- a/rubin_scheduler/scheduler/detailers/dither_detailer.py
+++ b/rubin_scheduler/scheduler/detailers/dither_detailer.py
@@ -361,6 +361,7 @@ class CameraSmallRotPerObservationListDetailer(BaseDetailer):
             # (we'll wipe this when we get to next fchange_idx)
             offsets[fchange_idx:] += self.per_visit_rot * np.arange(len(filter_list) - fchange_idx)
 
+		offsets = np.where(offsets > self.max_rot, max_rot, offsets)
         return offsets
 
     def __call__(self, observation_list, conditions):
--- a/rubin_scheduler/scheduler/detailers/dither_detailer.py
+++ b/rubin_scheduler/scheduler/detailers/dither_detailer.py
@@ -356,7 +356,7 @@ class CameraSmallRotPerObservationListDetailer(BaseDetailer):
         for fchange_idx, nvis_f in zip(filter_changes, nvis_per_filter):
             rot_range = self.rot_range - self.per_visit_rot * nvis_f
             # At the filter change spot, update to random offset
-            offsets[fchange_idx] = rng.random() * rot_range + self.min_rot
+            offsets[fchange_idx:] = rng.random() * rot_range + self.min_rot
             # After the filter change point, add incremental rotation
             # (we'll wipe this when we get to next fchange_idx)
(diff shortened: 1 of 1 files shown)


yoachim (approved):
Not sure why this is `CameraSmallRotPerObservationListDetailer` and not just `CameraSmallRotPerObservationDetailer`. 

Might think about if this should be setting `rotSkyPos_desired` and `rotTelPos_backup` rather than straight to `rotSkyPos` and `rotTelPos`.

Could probably use a unit test.

The `per_visit_rot` kwarg seems scary, like it could easily set offsets beyond `max_rot`. Seems like `self.max_rot` gets set, but never used to ensure things don't go beyond it.

bechtol (approved):
Quick functionality test

```
import numpy as np
from astropy.time import Time
from rubin_scheduler.scheduler.detailers import CameraSmallRotPerObservationListDetailer

detailer = CameraSmallRotPerObservationListDetailer(max_rot=45.0, min_rot=-45.0, seed=42, per_visit_rot=0.0)

filter_list = np.concatenate([np.tile('r', 5), np.tile('i', 5), np.tile('z', 5)])
mjd = Time('2024-11-18 01:00:00.000', format='iso').mjd 
initial_offset = -13.0

print(np.degrees(detailer._generate_offsets_filter_change(filter_list, mjd, initial_offset)))
```

Produces output

```
[26.56817209 26.56817209 26.56817209 26.56817209 26.56817209 36.81417199
 36.81417199 36.81417199 36.81417199 36.81417199 -2.1227076  -2.1227076
 -2.1227076  -2.1227076  -2.1227076 ]
```

Appears to be working

rhiannonlynne:
This is why another pair of eyes is good. I'll double check because I thought it was not going beyond max rot ..
Keith tested this for commissioning sv surveys but we found some problems last night so a unit test probably would be good.

bechtol:
> This is why another pair of eyes is good. I'll double check because I thought it was not going beyond max rot .. Keith tested this for commissioning sv surveys but we found some problems last night so a unit test probably would be good.
> 
> Going past max_rot wouldn’t be good, but there are two options: cutting off the dithers per visit, or reducing the per-visit dither so that it doesn’t pass max_rot .. not obvious to me which is preferable. @bechtol ? (in practice this hasn’t happened because the per-visit rot is very small or 0 and the rotator range is usually wide .. but can’t guarantee this is an always-case).

We should definitely put in a check. I do not have a strong preference on the implementation at this time. Would need to think about it more.

rhiannonlynne:
Ok for now I just added a cap for max_rot, so we don't go over the max value. 
Peter is correct that this should probably be setting rotTelPos_backup and rotSkyPos_desired rather than rotSkyPos directly. 
For now, I'm not sure it makes much difference and I don't know how they interact in the summit environment anyway.

rhiannonlynne:
going to merge and tag, see how long the rest of the process takes.

rhiannonlynne:
And of course should not have done that ... sigh, will try to figure out how to fix that mess.

rhiannonlynne:
Reopened ticket and made new PR to fix formatting issue.

PR #129, opened 2024-11-18
OPSIM-1209: Add rotation converter for comcam

rhiannonlynne:
Rotation converter for opsim, based on observed rotator and skyangle values in comcam on-sky data.


edennihy (approved):
Approved.

PR #131, opened 2024-11-19
remove scripted_id

yoachim:
Remove `scripted_id` from the observation schema. Have surveys append an index to the `scheduler_note` string so scripted surveys can still properly match incoming completed observations. Add a unit test to make sure scripted surveys are marking observations completed correctly.


rhiannonlynne (approved):

rhiannonlynne:
A curiosity that may or may not be relevant -- it looks like this means that things like DDF visits are tracked with the field name + filter of the visit ... is that right? is there also an integer? 
Separately, would there be an issue if the filter wasn't just 'ugrizy' but was something like 'r_03' ?

yoachim:
> A curiosity that may or may not be relevant -- it looks like this means that things like DDF visits are tracked with the field name + filter of the visit ... is that right? is there also an integer? Separately, would there be an issue if the filter wasn't just 'ugrizy' but was something like 'r_03' ?

The DDFs get an integer added to their scheduler_note field as well. I'll add a check to make extra sure no one can sneak around and set it up with non-unique note+filter combo in the script.

Should be fine with an "r_03" filter name. Hopefully we didn't cut corners and have "ugrizy" where we should have "["u", "g", ...]. Maybe a few default kwargs that should be updated to make it clear it should really be a list of str.

PR #139, opened 2024-12-02
SP-1763: Add predicted zeropoint utilities

rhiannonlynne:
Interpolation of zeropoint with airmass here: https://github.com/lsst-pst/syseng_throughputs/blob/main/notebooks/InterpolateZeropoint.ipynb
and then the resulting function is implemented here.

### DIFF
--- a/rubin_scheduler/utils/m5_flat_sed.py
+++ b/rubin_scheduler/utils/m5_flat_sed.py
@@ -1,4 +1,13 @@
-__all__ = ("m5_flat_sed", "m5_scale")
+__all__ = (
+    "m5_flat_sed",
+    "m5_scale",
+    "predicted_zeropoint",
+    "predicted_zeropoint_hardware",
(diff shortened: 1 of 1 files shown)


yoachim (approved):
Looks good.

Would be good to put in a link to where the fits were generated so it's easy to find if they need to be updated. 

Would be nice to have a token unit test. Maybe just loop over each function, band, airmass of [1.1, 2.5] and exptimes of [1,30,120] and assert the results are reasonable (like between 20 and 30).

rhiannonlynne:
My token unit test was even more token than that :) 
But ok, given that Tiago has preferences to keep these kinds of utilities out of rubin_scheduler, I'll move this over to rubin_sim instead. 
It sounds like he'd like us to move things like the consdb reading over to rubin_sim instead, and I can see that. 
Then the only thing rubin_scheduler should be doing would be converting between observations and opsim.

PR #140, opened 2024-12-03
add tolist method and obsarray_concat function

yoachim:
Add some methods/functions so it's easier to avoid loops in `Detailer` objects.


rhiannonlynne (approved):
Looks fine to me. 
I can see why this is easier to use for applying to larger sequences of visits. 
Does it add much overhead?

PR #142, opened 2024-12-09
Tickets/sp 1781

yoachim:
Update the `Survey` and `Detailer` API so they output `ObservationArray` objects of arbitrary length rather than a list of 1-element `ObservationArray` objects.


rhiannonlynne (approved):
I suppose this one merges first, since at least some of the other tickets look like they branched from this.

PR #146, opened 2025-01-06
Tickets/sp 1793

yoachim:
Swap "filter" kwargs, variable names, etc to "band". The `ModelObservatory` is updated to by default will in the "filter" field with the requested "band", so downstream things in MAF do not need updates.


rhiannonlynne (approved):
Looks good!

PR #147, opened 2025-01-14
arbitrary dither pattern detailer

yoachim:
Detailer to support adding arbitrary dither positions to a selected pointing.


rhiannonlynne (approved):
Looks great. Preserves the order of a sequence in a reasonable way too.

PR #149, opened 2025-01-19
SP-1841: Backwards compatible updates for Band/Filter

rhiannonlynne:
This PR has the goal to make the move from "filter" to "band" backwards compatible. 
This lets us tag at 3.5.0, making the band update a "feature" rather than breaking change, and lets @tribeiro deploy this at the summit without having to make over ts_scheduler. 
I've tested this with the current develop of ts_fbs_utils and the unit tests there run. I've also tested it with sim_runner and our baseline survey configuration. 
I am unsure of how to test this in a full run with ts_scheduler to ensure there are no breakages (this is technically true for any update we make). 

It would be good if both @tribeiro and @yoachim could take a look. 
The main place I am concerned about is in the Conditions object and in add_observation.
* in the Conditions: ts_scheduler currently reaches into the conditions object and adds information to relevant fields one-by-one. In this call, the "mounted_filters" and "current_filter" are set, using the physical filternames. When using the modelobservatory, both the current_band and current_filters are set directly using the band name.
With this PR, I have changed mounted_filters and current_filter to be set using a setter, so that if current_filter is set, then current_band is updated (splitting physical filtername down to band) and likewise for mounted_filters. Thus setting mounted_filters or current_filter will update the mounted_bands and current_band (which are the values which are then used inside the remainder of the FBS when referring to conditions values). 

* in add_observation: I am assuming that incoming observations from ts_scheduler will have 'filter' filled out (which could even be the full physical filter name) and will have 'band' as a field available to be filled, but that it won't have any information -- so then add_observation has a small "feature" at present to fill out 'band' if it wasn't already present (and should, in general split physical filter names properly down to band).  With sim_runner, 'band' will already be filled out and so nothing will happen.

### DIFF
--- a/rubin_scheduler/scheduler/features/features.py
+++ b/rubin_scheduler/scheduler/features/features.py
@@ -182,6 +182,8 @@ class NObsCount(BaseSurveyFeature):
         if self.scheduler_note == "":
             self.scheduler_note = None
         self.bandname = bandname
+        # Will not add filtername backup here; users should not reach
+        # down as far as the features in configurations.
 
     def add_observations_array(self, observations_array, observations_hpid):
         if self.scheduler_note is None and self.bandname is None:
--- a/rubin_scheduler/scheduler/basis_functions/basis_functions.py
+++ b/rubin_scheduler/scheduler/basis_functions/basis_functions.py
@@ -9,7 +9,9 @@ __all__ = (
     "M5DiffBasisFunction",
     "M5DiffAtHpixBasisFunction",
     "StrictBandBasisFunction",
+    "StrictFilterBasisFunction",
     "BandChangeBasisFunction",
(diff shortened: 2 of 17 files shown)


rubin_scheduler/scheduler/features/features.py
```diff
@@ -182,6 +182,8 @@ def __init__(self, scheduler_note=None, bandname=None, note=None):
         if self.scheduler_note == "":
             self.scheduler_note = None
         self.bandname = bandname
+        # Will not add filtername backup here; users should reach
+        # down as far as the features in configurations
```
yoachim, line 186: Maybe a missing "not" in here? Should be something like "users do not set feature.__init__ kwargs when configuring a scheduler obejct"

PR #150, opened 2025-01-21
Correct pairs to minimize filter changes

yoachim:
Making sure pairs are labeled and executed in the proper order.


ehneilsen (approved):
Looks good to me.

PR #151, opened 2025-01-23
refactor ddf_presched to make config easier. eliminate magic numbers

yoachim:
Refactor DDF schedulign code a bit in an attempt to make it a little easier for others to come in and use it. 

* Made `optimize_ddf_times` the only spot where scheduling logic parameters are set
* Removed magic numbers and COSMOS specific bits from `ddf_slopes` (this does make it return results that are not exact matches to previous versions)
* Converted `min_season_length` kwarg to be in days rather than a unitless fraction.


ehneilsen (approved):
Looks fine to me.

PR #152, opened 2025-01-24
SP-1865: Add filter backwards-compatibility to models

rhiannonlynne:
The previous pr for backwards compatibility missed filter/band swaps in the Models, so this is corrected here.

### DIFF
--- a/rubin_scheduler/scheduler/basis_functions/basis_functions.py
+++ b/rubin_scheduler/scheduler/basis_functions/basis_functions.py
@@ -13,6 +13,7 @@ __all__ = (
     "BandChangeBasisFunction",
     "FilterChangeBasisFunction",
     "SlewtimeBasisFunction",
+    "SlewtimeNoBandBasisFunction",
     "CadenceEnhanceBasisFunction",
     "CadenceEnhanceTrapezoidBasisFunction",
     "AzimuthBasisFunction",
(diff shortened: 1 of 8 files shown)


rubin_scheduler/scheduler/basis_functions/basis_functions.py
```diff
@@ -970,6 +971,47 @@ def _calc_value(self, conditions, indx=None):
         return result
 
 
+class SlewtimeNoBandBasisFunction(BaseBasisFunction):
```
yoachim, line 974: Why not just have `SlewtimeBasisFunction` take `None` as an option on the `bandname`?

rubin_scheduler/scheduler/surveys/field_survey.py
```diff
@@ -172,8 +174,15 @@ def __init__(
...
 
-        # Let's just make this an array for ease of use
-        self.observations = np.concatenate(self.observations)
+        # Let's just make this an array for ease of use if not already
+        if not isinstance(self.observations, ObservationArray):
+            self.observations = np.concatenate(self.observations)
+        # and backfill "band" if it was set by the user in "filter"
+        # Need to add "band" here if it wasn't populated
+        missing_band = np.where(self.observations["band"] == "")
+        band = np.char.rstrip(self.observations["filter"][missing_band], chars="_0123456789")
```
yoachim, line 183: This looks like hard-coding filter naming conventions that aren't documented anywhere. Should there be a utility function for converting `filter` to `band` so it's not a burried auto magical thing?

rubin_scheduler/scheduler/basis_functions/basis_functions.py
rhiannonlynne: I was going to do that .. I just figured this was enough of a change that it would be worthwhile to be REALLY clear that it was different than the basic SlewtimeBasisFunction. 
If you think it would be ok, I'm happy to change it.

rubin_scheduler/scheduler/surveys/field_survey.py
rhiannonlynne: Not a bad idea.

rubin_scheduler/scheduler/basis_functions/basis_functions.py
yoachim: I think the standard on most of the `Features` is to have a band set, and then if it is `None` it takes any band. So might as well have the same convention here.

rhiannonlynne: Done.

rubin_scheduler/scheduler/surveys/field_survey.py
rhiannonlynne: Ok, so because of how I needed to use this here, I added a utility that operates on the entire ObservationArray. I think this should only be needed for cases where sequences are set up in the configuration for FieldSurvey (and this should only be the case in configurations for the FBS which are not properly updated for the band-filter swap). 
The best documentation I could find about the filters for LSST was in fact in the code for obs_lsst, at https://github.com/lsst/obs_lsst/blob/main/python/lsst/obs/lsst/filters.py 
.. but you can see in the ConsDB for example "band" and "physical filter" names for many observations. 
Examples include r_03 == r.  For Auxtel, where gratings as well as filter can be used, this  particular conversion doesn't really work.

PR #154, opened 2025-02-04
refactor to eliminate duplicate kwargs min_area and area_required

yoachim:
Consolidate some repeated code into new method `calc_reward_basic`. Make sure the tests for if a survey has enough area are only being handled in one place with one relevant kwarg.


rhiannonlynne (approved):
Looks fine. 
It looks to me that this is backward compatible with existing surveys - could you please double-check by checking out ts_fbs_utils and running the unit tests in that repo against this version of rubin_scheduler? 
(this is probably a good thing to add to our unit test CI here, I'll work on that but for now. could you just check by hand?)

PR #155, opened 2025-02-20
updates to align more with DM style guide

humnaawan:


rhiannonlynne (changes requested):
Problems with "is" vs. an element-wise comparison, but once those are fixed it looks like.

humnaawan:
@rhiannonlynne I replaced the two instances of the not-element-wise checks - please lmk what you think.

also, would you know how to fix the CI check with the docs?

rhiannonlynne (approved):

PR #156, opened 2025-02-20
SP-1865: Additional changes for use with ts_scheduler v2.3

rhiannonlynne:
Missed that ts_scheduler was using defaults for SeeingModel setup, so didn't trigger the setting of self.filter_list.

### DIFF
--- a/rubin_scheduler/scheduler/detailers/detailer.py
+++ b/rubin_scheduler/scheduler/detailers/detailer.py
@@ -155,14 +155,10 @@ class BandToFilterDetailer(BaseDetailer):
     def __call__(self, observation_array, conditions):
         u_bands = np.unique(observation_array["band"])
         for band in u_bands:
-            indx = np.where(observation_array["band"] == band)[0]
-            if band not in self.band_to_filter_dict.keys():
-                # If the BandToFilterDetailer was set up with an
-                # empty dictionary (not None), then just copy the configured
-                # (possibly physical filtername) into band.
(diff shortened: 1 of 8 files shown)


rhiannonlynne:
Needed to also add 'note' back into observations sent to ts_scheduler (which does not yet know about or look at scheduler_note at that point). Also copied the content of scheduler_note into "note" so that this will return to the Target event log messages. 
I stripped "note" back out in SchemaConverter, so it is still not part of the official output databases, so the science community still should proceed in using scheduler_note.

rhiannonlynne:
Ok, and "filter" wasn't being filled in, with the older configurations that Tiago has. To avoid needing to add BandToFilterDetailers to all of the existing configurations for the FBS, I added BandToFilterDetailer as a *necessary* detailer for all surveys (in the BaseSurvey) class, and removed this mapping from ModelObservatory (as it is then unnecessary). 
The default setup for BandToFilterDetailer means that the band (even if it was physical filter name) will be copied into filter. 
Of course, physical filter name configurations will fail in simulations but may work on the summit with ts_scheduler. 
In general, we should strongly discourage configurations which use physical filter name going forward -- using "band" to configure, and then appropriately setting up the BandToFilterDetailer to map band into physical filter name is preferable and will allow use of the same configs in simulations and on summit.

rhiannonlynne:
Tested by Tiago - will consider that approval to progress to kicking off tag.

PR #157, opened 2025-02-25
euclid dither fix

yoachim:
Update euclid detailer to use latest scheduler_note convention.


ehneilsen (approved):
Looks fine to me, or at least I don't have any suggestions for improvement.
I'm a little worried about our general approach to identifying DDF fields by searching for substrings, but haven't thought of a better approach without schema changes that aren't more trouble than they're worth, so I'm not suggesting that you change it.
(What I'm worried about is that `scheduler_note` is a string, so it can technically have values like `10 degrees north of EDFS_a`, which is not something we want to identify as an EDFS field.)

yoachim:
That's a good point. It is easy to make it a little more specific to try and avoid that at least, so I'll do that.

PR #158, opened 2025-02-26
move band to filter mapping to CoreScheduler

yoachim:
Subclasses of the `Survey` object do not always run the base `__init__` method, so can fail to map band to filter.


rhiannonlynne (approved):

rhiannonlynne:
Discussed in slack. 
This seems fine, will require a new tag asap. 
Please do go ahead and add deprecation warnings to the unused too surveys if they're are not to be used.

PR #160, opened 2025-03-06
Move to newer setuptools_scm definitions

rhiannonlynne:


ehneilsen (approved):
Fine, as far as I can tell.

PR #161, opened 2025-03-07
SP-1686: Add a subclass of rubin_scheduler.utils.consdb.ConsDBVisits to support latiss

ehneilsen:
The code added assumes that the units of sky_db are ADU. It isn't presently filled out and I don't know the gain anyway, so the sky column returned ends up nan anyway, but what it should be should be checked.
Similarly, I have assumed the use of zero point I used for comcam will apply to latiss as well.
(I'm not 100% sure what I have there works even for comcam any more, because I recall discussion of converting things from ADU to electrons after I wrote this, and I don't know if the usage in consdb changed or not.)

### DIFF
--- a/rubin_scheduler/utils/consdb.py
+++ b/rubin_scheduler/utils/consdb.py
@@ -30,11 +30,11 @@ from rubin_scheduler.utils import (
 )
 
 GAUSSIAN_FWHM_OVER_SIGMA: float = 2.0 * sqrt(2.0 * log(2.0))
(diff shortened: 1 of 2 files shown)


rhiannonlynne:
I was curious - will this work outside of the USDF RSP? 
I don't know what the url is for the summit, although I assume they have an equivalent postgres server that could be queried. 
But outside of the USDF or the summit?

rhiannonlynne (approved):

ehneilsen:
> I was curious - will this work outside of the USDF RSP?

I don't know. I created a new Jira issue (SP-2013) to make sure it does.

PR #163, opened 2025-03-10
CloudMap

yoachim:
New object to hold cloud transparency maps. Updates to the `Conditions` and `M5DiffBasisFunction` to use the clouds.


rubin_scheduler/scheduler/features/conditions.py
```diff
@@ -85,8 +85,8 @@ def __init__(
         bulk_cloud : `float`
             The fraction of sky covered by clouds. (In the future might
             update to transparency map)
```
ehneilsen, line 87: This description might need updating?
I'm guessing that cloud_maps vs. bulk_cloud is either/or right now, but we still want to support both. It might be good to add some more explicit documentation of this.

PR #164, opened 2025-03-12
add mergesort kwarg to np.argsort

yoachim:
Set kwarg so numpy.argsort is cross-version and cross-platform stable.


ehneilsen (approved):
Looks fine to me.

PR #168, opened 2025-03-21
tickets/SP-2046: Quiet the benign "RuntimeWarning: All-NaN axis encountered" warnings when rubin_scheduler is run with reward recording turned on

ehneilsen:
Logs from pre-night sims are being flooded with these useless warnings. Get rid of them.


yoachim (approved):
yup, makes sense

PR #169, opened 2025-03-21
Tickets/sp 2037

yoachim:
Update spatial and rotational dithering so dither positions can changed per night, per call, or per visit.

### DIFF
--- a/rubin_scheduler/scheduler/detailers/dither_detailer.py
+++ b/rubin_scheduler/scheduler/detailers/dither_detailer.py
@@ -7,6 +7,8 @@ __all__ = (
     "DeltaCoordDitherDetailer",
 )
 
+import warnings
+
 import numpy as np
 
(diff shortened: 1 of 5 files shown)


rhiannonlynne:
I think this looks good and am approving the PR. 
I do have a question though: 
in the case of the CoreScheduler and surveys being instantiated at the start of each night, and then having the previous observations read back in, to get to the starting point for the night -- I don't see how the 'night' or 'call' (depending on which one is relevant) get updated to the expected starting point. I also don't see how you guarantee that the index value doesn't run over the max range of the indexes available (especially if you were using 'call' for an extended period of time).

yoachim:
Yes, now it should be possible to re-instantiate at the start of the night and have that be the same as running non-stop. I don't see an easy way to make it possible to re-build the scheduler in the middle of the night and have perfect repeatability. In that case the easiest thing is probably to just set new seeds for objects that dither.

PR #171, opened 2025-03-26
add AltAzShadowTimeLimitedBasisFunction

yoachim:
New basis function that can generate a mask in alt and az that only turns on at certain points of the night. 

Very open to suggestions for a shorter class name.


bechtol (approved):
I have not directly tested yet, but the implementation appears to capture the intended logic and is flexible enough to address the expected use cases for limiting the pointing azimuth range near the end of night.

I like the descriptive class name, and would not object to `AltAzShadowTimeLimitedMaskBasisFunction` for consistency with most of the other class names in this module that contain `MaskBasisFunction`.

edennihy (approved):
Looks good, thanks for the quick turnaround!

PR #173, opened 2025-03-28
SP-2902: Remove gitpython and conda following sim_archive move

rhiannonlynne:
Remove gitpython and conda from requirements.txt and pyproject.toml after moving the module that used these into rubin-sim.


ehneilsen (approved):
looks fine to me

PR #175, opened 2025-04-04
copy column detailer

yoachim:
Detailer so one can copy values from one column to another, e.g., `scheduler_note` to `target_id`.


rhiannonlynne:
Hmm. I suppose there might be some value in making sure that the type is still the same .. 
I would have thought that copying scheduler_note into target_id ought to fail because target_id should be an integer? 
I was actually thinking that this might be a way to make scheduler_note visible in "observation_reason" if it's not being picked up elsewhere, but those two are the same types (both strings).

yoachim:
It does fail if you try to copy a string into an int or float. But it does let you send a number into a string column, which I think makes sense.

rhiannonlynne (approved):

PR #176, opened 2025-04-04
Tickets/sp 2105

yoachim:
Code useful for science validation surveys. `TagRadialDetailer` that let's one label visits if they are close to a certain spot on the sky. `PerFilterStep` that can be used for short duration surveys to factor in when filters will be loaded, preventing filters that are not loaded from "falling behind" and resulting in a spike of single-filter visits trying to catch up.

### DIFF
--- a/rubin_scheduler/scheduler/utils/footprints.py
+++ b/rubin_scheduler/scheduler/utils/footprints.py
@@ -16,6 +16,7 @@ __all__ = (
     "slice_wfd_indx",
     "slice_quad_galactic_cut",
     "make_rolling_footprints",
+    "PerFilterStep",
 )
 
 import warnings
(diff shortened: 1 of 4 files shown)


rubin_scheduler/scheduler/utils/footprints.py
```diff
@@ -589,3 +590,75 @@ def calc_norm_factor_array(goal_map, radius=1.75):
...
+      Suervey length in days. Default 80.
+    bands : `list`
+        Bandnames. Defualt of none uses ugrizy
+    loaded_dict : `dict`
+        Dict with keys of bandnames and values of
+        arrays of ints with the nights that band is
+        loaded. E.g.,
+        loaded_dict = {'u': np.array([1,2,3,15,16,17])}
+        means the u band will only be availble on those
+        nights.
```
rhiannonlynne, line 614: Does this kwarg mean that the times that filters will be loaded over the SV surveys has to be known at the start of the SV surveys, in order for this to work properly?

rhiannonlynne (commented):
I'll review again when unit tests are finished, but have a comment.

rubin_scheduler/scheduler/utils/footprints.py
rhiannonlynne: I'm guessing we may not know this exactly (especially for the earlier periods), but maybe getting close enough is fine.

yoachim: Just knowing a filter will be removed for some time in the future should be enough to prevent a big spike in observations.

PR #177, opened 2025-04-14
detailers for target name

yoachim:
Detailers for labeling which region a visit is in and if it is close enough to be included in DDF observations. Also a small update so `Detailer` sub-classes don't need to bother defining an unused `survey_features` dictionary.

### DIFF
--- a/rubin_scheduler/scheduler/detailers/detailer.py
+++ b/rubin_scheduler/scheduler/detailers/detailer.py
@@ -20,21 +20,25 @@ __all__ = (
     "BandToFilterDetailer",
     "TagRadialDetailer",
     "CopyValueDetailer",
+    "LabelRegionDetailer",
+    "LabelDDFDetailer",
 )
 
 import copy
 import warnings
 
+import healpy as hp
 import numpy as np
 
 import rubin_scheduler.scheduler.features as features
-from rubin_scheduler.scheduler.utils import IntRounded, ObservationArray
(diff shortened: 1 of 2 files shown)


rubin_scheduler/scheduler/detailers/detailer.py
```diff
@@ -683,6 +688,117 @@ def __call__(self, obsarray, conditions):
...
+            labels = np.unique(self.label_array[indx])
+            # ignore things that are out of bounds
+            labels = labels[np.where(labels != "")]
+            # If we overlap multiple regions
+            if np.size(labels) > 1:
+                result = ", ".join(labels)
+            else:
+                result = labels[0]
+
+            if self.append:
```
rhiannonlynne, line 737: How would append come to be used? Isn't an observation only proposed from a single survey, so it should only go through the detailer once?

rhiannonlynne (approved):
Look good. 
Would it make sense to cache the lookup from RA/Dec to labels, or do fields just not get repeated very often?

rubin_scheduler/scheduler/detailers/detailer.py
yoachim: There can be multiple detailers (and they survey itself) setting the `target_name`. So a ToO in a DDF could end up with a target name like `SSO_night_i1, lowdust, EDFS_a`.

PR #178, opened 2025-04-16
Turn down IntRounded warning messages, convert to int only if needed

yoachim:
Make `IntRounded` smarter so it only converts values to ints if there's a potential that a comparison will be near the machine-precision limit.

### DIFF
--- a/rubin_scheduler/scheduler/utils/utils.py
+++ b/rubin_scheduler/scheduler/utils/utils.py
@@ -25,6 +25,7 @@ __all__ = (
     "rotx",
 )
 
+import copy
 import datetime
 import os
 import socket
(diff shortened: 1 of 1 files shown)


rubin_scheduler/scheduler/utils/utils.py
```diff
@@ -153,50 +154,97 @@ def mean_azimuth(az, min_val=0.1):
...
+        Some number/array that we will want to compare
+    scale : `float`
         How much to scale inval before rounding and converting to an int.
+        Essentially how many digits of precesion to use in comparisons.
+        Default 1e6.
+    limit : `float`
+        If values have a differce less than the limit, force comparisions
+        to be done with scaled ints. Should be larger than whatever you
+        think the machine floating point precision might be.
+        Default 1e-9.
```
ehneilsen, line 181: Why not use numpy.finfo(float) (or whatever type you start with) to get this automatically?

ehneilsen (approved):
I suggest setting `limit` to `numpy.finfo(float).resolution` by default (or maybe multiply it by 10 to be safe).
If you wanted to be more precise about it, you could rework it to use numpy.nextafter instead of using limit, but that's probably more trouble than it's worth at this point.

rubin_scheduler/scheduler/utils/utils.py
yoachim: 1) I'm not sure I believe `numpy.finfo` because it gives me the same value on systems where float comparisons behave differently (I think)
2) We need the default limit to be above the limit on the lower resolution machine. So auto setting using  `numpy.finfo` means we could be comparing native values that would be rounded to lower precision on different hardware.

PR #179, opened 2025-05-06
ToO maps can be any nside

yoachim:
Make sure a footprint set by a ToO event object can be any nside and `Survey` object will change resolution if needed.


ehneilsen (approved):
looks fine

PR #180, opened 2025-05-07
Adding different BBH ToO surveys

seanmacb:


seanmacb:
The above changes are derived from [here, page 21](https://arxiv.org/pdf/2411.04793)

> Area: Based on our O5 simulations, we consider that 3 events may be covered with only one
Rubin pointing, 2 would require 2 pointings, and 2 would require 3 pointings. With 30s
exposures for 3 filters, this would take 5 epochs, for 7 events. With 30s exposures for 3 filters,
and 5 epochs, for 7 events we recommend a total of 4.3 hours per year, including overheads.
For the rare case that a candidate strongly lensed BBH is detected, we will select it based on the
detection of multiple GW images as multiple (at least 2) significant events, and specifically on
their sky maps overlap statistics (Wong et al., 2021) between multiple skymaps having a false
positive rate < 10^-2. The overlapping skymap is what we would follow up on, and it has to
satisfy the requirement of 𝛺 < 20 sq. deg. as for other BBH events, but we do not require it to
fall within the Rubin footprint, given the rarity of such an event. We only recommend one trigger
of this kind over the entirety of O5, should it occur, it will be considered as part of the total time
recommendation above.
Filters: If it is not bright time, we recommend u- and g-band as several BBH in AGN models
predict a blue fast component and also we want to discern them from TDEs. TDE studies show
that u-band rest frame is ideal for identifying these objects. At the typical distances of these
events, u or g should cover the rest frame u-band. We suggest the use of ugi for events at z<0.4
and gri for more distant events as baseline and switch to riz in bright time.
Timing: The first epoch should be taken as soon as possible to get a baseline - the AGN should
not be flaring at that time yet: if the merger occurred within the disk photosphere, it will emerge
over a timescale which is around the photon diffusion timescale. This timescale for emission is
expected to be highly dependent on the binary properties and the environment, hence we propose
to observe on a log-spaced timescale as proposed in the BBH case presented in Andreoni et al.,
2022, e.g. days 1-3-8-10-40 after trigger. We will complement ToO observations with nominal
cadence LSST observations. Except for the first epoch, if regular survey observations occur in
the region of the GW event in one or more of the proposed filters the day before the planned ToO
epoch, that ToO epoch can be skipped, hence our time recommendation can be considered
conservative. Longer (>40 days) timescales (which may be relevant for the events at the largest
21
distances) will be more likely to be probed by nominal survey cadence, hence we do not propose
for those.

seanmacb:
The lensed BNS change is from [here, page 16](https://arxiv.org/pdf/2411.04793):

>Area: For a representative area of Ω = 150 degree2, we would reach depths of g ≈ 26.2 and r ≈
26. We implement our “as deep as possible” approach by adopting a fixed ~5 hours of elapsed
time per epoch, with the depth that can be achieved in this time being a function of GW sky
localization uncertainty. Actual ToO triggers will fall in the range bracketed by the following two
scenarios which are also shown in Figure 3. Note that smaller GW sky localization uncertainties
allow us to push to fainter photometric limits and test fainter kilonova models within our fixed
envelope of 5 hours per epoch:
Ω = 900 degree2: This is the threshold at which we can go deep enough in ~5 hours per
epoch to test the lensing hypothesis for the AT170817-like model. We estimate that this
would require 30 seconds per pointing in g-band and 90 seconds per pointing in r-band,
to reach depths of g = 24.8 and r = 25.
Ω = 15 degree2: Over this area, we can go deep enough in ~5 hours per epoch to test the
lensing hypothesis for a wide range of kilonova models including the conservative model
shown in Figure 3. This should require ~1.5 hours of integration time per pointing in
g-band and 1 hour of integration time per pointing in r-band, to reach depths of g = 27.6
and r = 27.

seanmacb:
The neutrino change is derived from [here, page 31](https://arxiv.org/pdf/2411.04793):
 
> Day 0:
> We suggest beginning observations ASAP after neutrino localization, with a single deep
> observation in g-band. (120s)
> We also recommend r-band (30s) separated at least 15 minutes from g-band, and z-band
> observations (30s) at some point in the night. We do not have strict conditions for the
> timing of the z-band observations. r-band is important to provide a second detection for
> most transients, z-band to identify extremely red sources (e.g. GRB afterglows).
> Day 1:
> We would again observe in g-band (120s), and r-band (30s). This will resolve rapidly
> evolving transients, sensitive both to fading/rising and color change.
> Day ~7 (with some flexibility to minimize survey disruption):
> We would observe one final time in g (120s), r (30s), and z (30s) to detect a wide variety
> of transients that may have evolved since day 0 in color or luminosity.
> Day X
> We additionally recommend one u-band observation (30s) for color information, either on
> day 0 or whenever the u-band filter is next mounted. u-band is a powerful method to
> identify TDEs/LFBOTs, and we expect a typical median latency of ~1 week.

yoachim (approved):
Looks good. I fixed an issue where some 5's became 3's, hopefully made it more clear which things need to match the length of `times`.

seanmacb:
Thanks Peter, this makes it very clear. I will merge this to main.

PR #181, opened 2025-05-07
U/yoachim/too events detailers

yoachim:
Add new `Detailer` subclass to help dynamically adjust ToO scripted visits.

### DIFF
--- a/rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
+++ b/rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
@@ -6,7 +6,7 @@ from copy import copy
 import healpy as hp
 import numpy as np
 
-from rubin_scheduler.detailers import BandPickToODetailer
+from rubin_scheduler.scheduler.detailers import BandPickToODetailer
 from rubin_scheduler.scheduler.surveys import BaseMarkovSurvey, ScriptedSurvey
 from rubin_scheduler.scheduler.utils import (
     ScheduledObservationArray,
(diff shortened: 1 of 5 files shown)


rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
```diff
@@ -495,6 +509,10 @@ def gen_too_surveys(
         )
     )
 
+    event_detailers = [
+        BandPickToODetailer(band_start="r", band_end="i", distance_limit=10.0, check_mounted=True),
+        BandPickToODetailer(band_start="i", band_end="z", distance_limit=100.0, check_mounted=True),
```
seanmacb, line 514: I will verify that this distance we want this transition at ... I am not finding it currently

rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
seanmacb, line 1: An additional place we want the detailers is for the BBH events. The filters follow this kind of structure:

- For BBH events in dark time (u mounted) and mean posterior distance < 2200 Mpc, use filters ugi
- For BBH events in dark time and mean posterior distance > 2200 Mpc, use filters gri
- For BBH events in bright time, use filters riz

Exposure times, visit numbers, and visit times should stay the same

seanmacb (changes requested):
The only change is to add the detailer for the BBH events (I can also do it)

rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
seanmacb: After consultation with the team wanted to default to g+i, and could not commit to a distance threshold at this time.

seanmacb (approved):
Changes to support the dark time requirement look okay, but it would be good if someone else puts eyes on this

PR #183, opened 2025-05-16
add moon wind avoidance to ToOs and Roman

yoachim:
Make sure ToO and Roman surveys have lunar and wind avoidance.


rhiannonlynne (approved):
Looks like this should work. 
Do we have a test that checks both moon avoidance basis functions make the same kind of results? Mostly future-proofing in case we get better guidance on moon avoidance as a function of phase, elevation, etc.

PR #184, opened 2025-05-21
Seanmacb/to o 3strategies

seanmacb:
Two proposed changes here
- Increased flush time for ToO surveys - before this change, only 2 days are observed for most ToO events
- Added a galactic SN ToO case. This may need further development and testing, but the simple case is handled here now

### DIFF
--- a/rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
+++ b/rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
@@ -491,7 +491,7 @@ def gen_too_surveys(
             survey_name="ToO, GW_case_B_C",
             target_name_base="GW_case_B_C",
             split_long=split_long,
-            flushtime=48,
+            flushtime=80,
             n_snaps=long_exp_nsnaps,
         )
     )
@@ -518,7 +518,7 @@ def gen_too_surveys(
             survey_name="ToO, GW_case_D_E",
             target_name_base="GW_case_D_E",
             split_long=split_long,
-            flushtime=48,
+            flushtime=80,
             n_snaps=long_exp_nsnaps,
             event_gen_detailers=None,
         )
@@ -564,7 +564,7 @@ def gen_too_surveys(
             survey_name="ToO, BBH",
             target_name_base="BBH",
             split_long=split_long,
-            flushtime=48,
+            flushtime=40*24,
             n_snaps=n_snaps,
             event_gen_detailers=event_detailers,
         )
@@ -655,7 +655,7 @@ def gen_too_surveys(
             survey_name="ToO, neutrino",
             target_name_base="neutrino",
             split_long=split_long,
-            flushtime=8.0,
+            flushtime=150,
             n_snaps=n_snaps,
         )
     )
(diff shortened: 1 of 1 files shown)


seanmacb:
Also, correction to the GW silver strategy first night exposure times

rhiannonlynne:
I have a question (probably for @yoachim)
is there any logic in the ToO surveys so that if one of the sets of visits is delayed, that the follow-up also gets delayed? 
With a flush time of 80 hours, if the first visits are delayed by 80 hours this puts the first observations later than the last follow-up would otherwise have occurred. They won't all trigger at once, I assume? And if we do start 80 hours late than expected, are 80 hours + 72 hours (after the initial trigger) useful for the last set of visits?

rhiannonlynne:
And another question for @seanmacb - 
The implementation here for the galactic SN doesn't quite seem in line with what I think they actually want, which is just "look at this patch of the sky until we say stop, which is probably less than 48 hours". 
Were the 100 visits requested here intended to be a simplification in order to allow us to simulate these events?

seanmacb:
> The implementation here for the galactic SN doesn't quite seem in line with what I think they actually want, which is just "look at this patch of the sky until we say stop, which is probably less than 48 hours". Were the 100 visits requested here intended to be a simplification in order to allow us to simulate these events?
Yes, it was just a placeholder. The actual implementation may be a special scheduler config as you mentioned on slack, or a BLOCK, but this will allow me to run a few sims to verify that it produces _some strategy_ that follows the intention

seanmacb:
Many of the ToO cases provide an explicit mention of the scientific loss if day 1 observations are not acquired for any reason. For example, the galactic SN case:

> Scientific loss if day 1+ observations are not acquired: If there is a small delay between the neutrino trigger and shock breakout (<1 day), we could miss the rise of the SN as well as not be able to improve upon the neutrino localization

What is the best way to handle this, from an operational perspective? changing the flush time?

rhiannonlynne:
The flush time changes how long the FBS considers a requested scripted observation to be still valid. Increasing the flush time increases the time that the observation will still be attempted. In terms of delays .. if we can take the observation, we will, since these ToO observations are in the highest tier (the only way they would fail to be queued up is if there is another long running sequence blocking it .. the longest running sequences right now are other ToO sequences, I think). The only other way we would not choose those observations is if we can't -- if the field is not visible or the observatory is not operating (or we might have something set so that we don't choose ToOs outside of our survey footprint, or where we don't have templates .. but that's different). 

As such, I'm not sure we need to handle this question from an operational perspective .. we'll get the observation if we can, and if it's practical, and if not - we couldn't for what ought to be valid reasons.

rhiannonlynne:
(looks like you need to run black)

seanmacb:
Let me revert the flush time change - the 48hr case should be sufficient then...

yoachim (approved):
I set it so the Galactic SNe now alternate between 1 and 15s visits, cycling 8 times.

seanmacb:
I added minor changes to the neutrino, GW gold, and GW silver strategies. All verified from testing - see [here](https://github.com/seanmacb/LSST-ToO-SchedulerSims/blob/main/notebooks/too_test_event.py) for script of testing different strategies and creating relevant metrics for comparison.

PR #185, opened 2025-05-24
Adding changes to lensed BNS strategy from testing

seanmacb:
Minor changes to lensed BNS strategy, as a result of testing. See [slack](https://rubin-obs.slack.com/archives/C08Q47VJZJ7/p1748105087008959) for report on testing

### DIFF
--- a/rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
+++ b/rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
@@ -608,10 +608,10 @@ def gen_too_surveys(
     # Lensed BNS
     ############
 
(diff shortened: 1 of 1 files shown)


rhiannonlynne:
@yoachim I thought there might have been a reason we didn't include these later BNS observations into the survey before - can you comment please?

seanmacb:
I think long term, this strategy will be revised to have some sort of detailer according to the 90% region. Right now, we have the upper and lower ends of a 90% area, but there should be a sliding function (detailer) that can tell you that given a 90% are of X, what the exposure times, visits, etc.

This case can at least handle the upper and lower ends of the alert type

PR #187, opened 2025-05-29
Tickets/sp 2256

yoachim:
Basis functions that are helpful for `Survey` objects that are dedicated to gathering image difference templates.


rhiannonlynne (approved):
~I think this one just has to merge after your other branch, as it shares some commits.~  
all good, managed to delay long enough in approving it. 
But looks fine.

PR #189, opened 2025-06-09
resolve rotSkyPos if out of range in ModelObservatory

yoachim:
Previously, the scheduler set `rotSkyPos_desired` and `rotTelPos_backup` to ensure each observation could have a valid rotation angle set. Now, the scheduler sets only `rotSkyPos`, and if that value is invalid, `ModelObservatory` tries rotating the requested value by -90, 90, -180, and 180 degrees to find a valid rotation. This should more closely emulate what happens at the telescope.

### DIFF
--- a/rubin_scheduler/scheduler/model_observatory/model_observatory.py
+++ b/rubin_scheduler/scheduler/model_observatory/model_observatory.py
@@ -12,6 +12,7 @@ import rubin_scheduler.skybrightness_pre as sb
 from rubin_scheduler.data import data_versions
 from rubin_scheduler.scheduler.features import Conditions
 from rubin_scheduler.scheduler.model_observatory import KinemModel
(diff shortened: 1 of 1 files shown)


yoachim:
I don't know what the ts sequence is.

rhiannonlynne:
I think the default is this  -  [180.0, -180.0, 90.0, -90.0]
https://github.com/lsst-ts/ts_observatory_control/blob/develop/python/lsst/ts/observatory/control/base_tcs.py#L131
although it also look like it could be overriden if desired (I haven't seen any place where that override happens though).

rhiannonlynne:
Your strategy looks more interesting. 
Wonder if we could get it in lsst_ts instead.

rhiannonlynne (approved):
We'll leave it like this for now, and change if necessary. Hopefully small impact compared to summit.

PR #195, opened 2025-07-04
Update too_scripted_surveys.py to remove duplicate detailers call

seanmacb:
Very minor change, but a duplicate call of detailer definition on the ToO scripted survey is removed here.


yoachim (approved):
Good catch.

Be sure to request someone to review when the PR is ready (I assumed this was ready tor review so tagged myself)

PR #197, opened 2025-07-09
Tickets/sp 2417

yoachim:
Update `LongGapSurvey` so it doesn't use `rotSkyPos_desired` anymore. Make sure ToO surveys are using unique detailer lists.

### DIFF
--- a/rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
+++ b/rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
@@ -1,7 +1,7 @@
 __all__ = ("ToOScriptedSurvey", "gen_too_surveys")
 
 import warnings
-from copy import copy
+from copy import copy, deepcopy
 
 import healpy as hp
 import numpy as np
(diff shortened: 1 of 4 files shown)


rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
```diff
@@ -202,19 +202,17 @@ def __init__(
 
         # list to keep track of alerts we have already seen
         self.seen_alerts = []
-
         # Add information for visit metadata if necessary
+        has_tracking_detailer = False
+        for detailer in self.detailers:
+            if isinstance(detailer, TrackingInfoDetailer):
+                has_tracking_detailer = True
```
rhiannonlynne, line 209: Why move this check earlier and not break out of the loop if one is found? 
I am missing something about why the previous implementation didn't work.

rhiannonlynne (commented):
I understand and appreciate copying the detailer lists (ah, mutable arguments.. ). 
But I don't understand why the check on if the tracking detailer was present needed to change, and I'd like to. Can you clarify please?

rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
yoachim: I think this was just a readability change I made while debugging. Having a `break` so nested can be tough to parse.

PR #198, opened 2025-07-11
Update too_scripted_surveys.py to combine neutrino ToO surveys

seanmacb:
This change is to integrate the u band neutrino observations into the other neutrino ToO observations. This change will make the alert receipt easier, since there will no longer be a need for `neutrino_u` and `neutrino` alert types, which would be duplicate skymaps anyways.

The only place I am unclear about is if the flush time is sufficient, since we would likely need a longer (~30 days?) flush time to ensure we provide the opportunity for scheduled neutrino u-band observations.

### DIFF
--- a/rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
+++ b/rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
@@ -666,9 +666,10 @@ def gen_too_surveys(
     # Neutrino detector followup
     ############
 
-    times = [0, 15 / 60.0, 0, 24, 24, 144, 144]
-    bands_at_times = ["g", "r", "z", "g", "r", "g", "rz"]
+    times = [0, 0, 15 / 60.0, 0, 24, 24, 144, 144]
+    bands_at_times = ["u", "g", "r", "z", "g", "r", "g", "rz"]
     exptimes = [
+        30,
         120,
         DEFAULT_EXP_TIME,
         DEFAULT_EXP_TIME,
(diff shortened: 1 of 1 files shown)


rhiannonlynne:
The flush time should be able to be set differently for different observations in the set of scheduled_observations.
However, I'm not entirely certain if a 30 day flush time is appropriate either.
If we did not get observations quickly, wouldn't the object be too faint? 
And if we did get observations earlier and localized the ToO, then wouldn't followup happen with other telescopes to get the u band visits we could not?

seanmacb:
From the strategy definition paper

> There will be many sources detected in Rubin observations, and it won’t be feasible to perform follow-up for each. Day 1 observations are essential to identify rapidly-evolving sources, and Day 7 observations are needed to perform photometric classification. Only with the full sequence can promising candidates be feasibly selected for follow-up.

So night zero observations seem extremely important.

As for flush time, the only set of observations that require a long flush time are the u band observations. 

The u-band observations are critical for TDEs and fast blue optical transients. As for when the cutoff is, I will contact the neutrino specialist who devised the strategy

seanmacb:
Here's the answer I received

> The idea behind u band is that is an excellent way to distinguish between supernovae and Tidal Disruption Events. TDEs are relatively long lived, lasting months, and stay UV bright the whole way. So the observations remain valuable even if there is a delay of e.g a month. The one issue is that the further you get from the last g band observation, the harder it is to estimate what the u-g colour is. So if there was a delay of e.g 20 days, it would probably be better if there was u taken alongside an extra 30s g.

seanmacb:
More

>I haven't done a dedicated study, but my experience here is that within +/- 1 week there will not be much lightcurve change. There is a g visit on night 0, 1, and 7, so I'd say night 14 onwards would benefit from an additional g.

yoachim (approved):
Makes sense to me.

PR #200, opened 2025-07-11
SP-2420: Add utility function for special targets (Roman field)

yoachim:
Function to hold ""special locations" like Roman bulge field postion.

### DIFF
--- a/rubin_scheduler/utils/ddf_locations.py
+++ b/rubin_scheduler/utils/ddf_locations.py
@@ -1,7 +1,32 @@
-__all__ = ("ddf_locations_skycoord", "ddf_locations", "ddf_locations_pre3_5")
+__all__ = ("ddf_locations_skycoord", "ddf_locations", "ddf_locations_pre3_5", "special_locations")
 
 import astropy.units as u
 from astropy.coordinates import SkyCoord
+
+
+def special_locations(skycoords=False):
+    """Special locations.
+
+    Parameters
+    ----------
+    skycoords : `bool`
+        Return locations as astropy.SkyCoords. If False, returns
+        as a tuple of floats. Default False.
+    """
+
+    # Roman bulge field location
+    roman_ra_deg = 268.708
+    roman_dec_deg = -28.975
+
+    result = {}
+    if skycoords:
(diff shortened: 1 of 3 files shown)


rubin_scheduler/utils/ddf_locations.py
```diff
@@ -1,30 +1,62 @@
...
     ddf["ELAISS1"] = SkyCoord("00:37:48 −44:01:30", unit=(u.hourangle, u.deg), frame="icrs")
     ddf["XMM_LSS"] = SkyCoord("02:22:18  −04:49:00", unit=(u.hourangle, u.deg), frame="icrs")
     ddf["ECDFS"] = SkyCoord("03:31:55  −28:07:00", unit=(u.hourangle, u.deg), frame="icrs")
     ddf["COSMOS"] = SkyCoord("10:00:26  +02:14:01", unit=(u.hourangle, u.deg), frame="icrs")
     ddf["EDFS_a"] = SkyCoord(ra=58.90 * u.deg, dec=-49.32 * u.deg, frame="icrs")
     ddf["EDFS_b"] = SkyCoord(ra=63.60 * u.deg, dec=-47.60 * u.deg, frame="icrs")
-    return ddf
 
+    if skycoords:
+        return ddf
```
rhiannonlynne, line 57: So .. for readability's sake, rather than returning but then also having code after the return (which can be harder to read, but is worthwhile sometimes) ... 

would it be possible to just replace 56 - 62 with 

```
if not skycoord:
   # replace SkyCoord with ra/deg tuple in degrees
    ddf = dict([(n, (ddf[n].ra.deg, ddf[n].dec.deg)) for n in ddf])
 ```
 
 and then just return ddf? 
 (although I admit that's not the most readable dict / list comprehension, so 
 ```
 if not skycoord:
    result = {}
    for n in ddf:
        result[n] = (ddf[n].ra.deg, ddf[n].dec.deg)
    ddf = result
```
also seems reasonable.

rhiannonlynne (commented):
Because this PR was built on top of a different branch that is also under PR, it seems difficult to approve this one until the other one is finished ...

rhiannonlynne (approved):

PR #202, opened 2025-07-17
Update too_scripted_surveys.py to have one field per healpix pixel

seanmacb:
Here, I make an attempt to improve the fields for the ToOScriptedSurvey class.

This change gives each healpix pixel a single pointing. For our current `nside=32` implementation, this means a mean spacing of ~1.8 deg^2. While this is still not optimal, it is already an improvement ahead of the current implementation which has anywhere from 3-4 pointings within a single healpix pixel.

Longer term, each ToO event should have its own pointing grid created, but that is a larger conversation outside of this immediate scope

### DIFF
--- a/rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
+++ b/rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
@@ -16,8 +16,8 @@ from rubin_scheduler.scheduler.utils import (
     thetaphi2xyz,
     xyz2thetaphi,
 )
-# from rubin_scheduler.site_models import _read_fields # commenting out, following stopgap
-# solution for pointings
+# from rubin_scheduler.site_models import _read_fields # commenting 
+# out, following stopgap solution for pointings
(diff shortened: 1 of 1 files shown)


seanmacb:
The optimization I am trying to make here is the improvement to the pointing generation. Right now, we have 3-4 pointings in a given healpixel, which oversamples the regions we care about.

My intention with this change was to have one pointing centered at each healpixel, such that when we pass a map to the scheduler for a ToO, each healpix we request to observe is associated with one pointing, not 3-4.

If there is a better way to do this, let me know and I can work on it. This was intended as a stopgap measure, since longer term, each ToO event should have a unique pointing grid

yoachim:
I don't think the code currently generates 3-4 pointings per HEALpix, or if it does it's doing somewhere else in the code and not here. It looks to me like you're changing a sparse HEALpix to pointing mapping with a denser HEALPix to pointing mapping. That's why I think a unit test would be the way to go about fixing the issue.

I'll tag you to review a branch that might solve the problem.

PR #203, opened 2025-07-21
Tickets/sp 2467

yoachim:
* Update the mapping of HEALpix to pointing in general. 
* Add additional kwargs to help control ToO dithering behavior
* Have ToO pointing tessellation cover HEALpix centers AND all corners of a desired HEALpix.


seanmacb (approved):
Thanks for addressing this. This is what I had in mind. When this is merged, I will delete my PR.

rhiannonlynne:
This is definitely an improvement, and we should merge it.
However, it still does not completely address the problem of a well localized ToO because the RA/Dec of the object could have been in the edge of the healpix (and since we center on the center of the healpix, we're still non-centered on the ToO).
I think it also means that we're likely to use multiple pointings for a single healpix (still) to make sure we cover the entire healplix.  

I think we really need the packets to have no healpix points when they really only want a given ra/dec and know that the entire localization will be well within a single FOV, and then we just point at the ra/dec of the ToO. 

But since this is still an improvement, I think this is reasonable to merge now and then go from there.

PR #204, opened 2025-07-24
Update too_scripted_surveys.py to iterate over filters before visits

seanmacb:
In the current implementation, we iterate over filter before iterating over visits. For four pointings, with two visits, each in r and i band, that looks like 

> r1, r2, r3, r4, i1, i2, i3, i4, r1, r2, r3, r4, i1, i2, i3, i4

where the integer represents the pointing number. This increases the total number of filter changes, and is not optimal for maximizing the open shutter fraction. 

This PR aims to swap the iteration order, instead iterating over visits before iterating over filters. For the example above, the intended behavior is then 

> r1, r2, r3, r4, i1, i2, i3, i4, r1, r2, r3, r4, i1, i2, i3, i4

For most science cases, the usage of multiple visits is to coadd images to reach a desired depth before performing DIA. One example of this is the GW NS gold case, which requests 180/120s of exposure time for visit epochs, which cannot be accomplished by simply increasing the exposure time for a single visit due to saturation limits.

If this is not the best way to do this, please let me know.

### DIFF
--- a/rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
+++ b/rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
@@ -349,21 +349,21 @@ class ToOScriptedSurvey(ScriptedSurvey, BaseMarkovSurvey):
                     self.exptimes,
                     np.arange(np.size(self.times)),
                 ):
-                    for i in range(nv):
-                        # let's dither each pointing
-                        if (i != 0) & (hpid_to_observe.size > 0):
-                            ras, decs = self._tesselate(hpid_to_observe)
+                    for bandname in bandnames:
+                        # Subsitute y for z if needed on first observation
+                        if i == 0:
+                            if (bandname == "z") & (bandname not in conditions.mounted_bands):
+                                bandname = "y"
+    
+                        if bandname == "u":
+                            nexp = self.n_usnaps
+                        else:
+                            nexp = self.n_snaps
 
-                        for bandname in bandnames:
-                            # Subsitute y for z if needed on first observation
-                            if i == 0:
-                                if (bandname == "z") & (bandname not in conditions.mounted_bands):
-                                    bandname = "y"
-
-                            if bandname == "u":
-                                nexp = self.n_usnaps
-                            else:
-                                nexp = self.n_snaps
+                        for i in range(nv):
+                            # let's dither each pointing
+                            if (i != 0) & (hpid_to_observe.size > 0):
+                                ras, decs = self._tesselate(hpid_to_observe)
 
                             # If we are doing a short exposure
                             # need to be 1 snap for shutter limits


yoachim (approved):
This makes sense. I think this means one could put in ''grg" for the bands to get the old behavior if they wanted to

rhiannonlynne:
If there is a JIRA ticket associated with this, it's nice to add the JIRA ticket into the branch name and pull request.

rhiannonlynne:
My question I had for Sean about this PR was if the change in order mean that the dithering & calculation of the tessellation was more than was expected, since there are more calculations of the pointing + tessellation in this order than previously. 
Before the behavior was to pick an RA/Dec then cycle through the bands (so I think that means that each pointing would be covered with the same boresight in all bands) - Now the behavior is to cycle through the bands and pick the RA/Dec within the band (so I think that means that the boresight varies between bands). 
Maybe this is not important.

seanmacb:
Apologies for letting this slip

**Injected BBH event**

```
footprint = np.zeros(hp.nside2npix(32))
indices = hp.query_disc(32,hp.ang2vec(225,-50,lonlat=True),radius=2*np.pi/180)
footprint[indices]=1

myToO = TargetoO(
    1,
    footprint,
    60884,
    10,
    ra_rad_center=225*np.pi/180,
    dec_rad_center=-50*np.pi/180,
    too_type="GW_case_B",
)```

**Associated observing strategy**

```
bf_list = []
detailer_list = [detailers.DitherDetailer(
        per_night=True, max_dither=0.001
    )]
times = np.array([0, 2, 7, 9, 39]) * 24
bands_at_times = ["uri"] * times.size
nvis = [2] * times.size
exptimes = [30] * times.size

nside=32

# This is a duplicate of the BBH  strategy from
# rubin_scheduler/rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
too_survey = [
    ToOScriptedSurvey(
        bf_list,
        nside=nside,
        followup_footprint=np.ones(hp.nside2npix(nside)),
        times=times,
        bands_at_times=bands_at_times,
        nvis=nvis,
        exptimes=exptimes,
        detailers=detailer_list,
        too_types_to_follow=["GW_case_B", "GW_case_C"],
        survey_name="ToO, BBH, custom",
        target_name_base="ToO_BBH",
        split_long=False,
        flushtime=48,
        n_snaps=1,
        science_program="BLOCK-123",
    )
]
```

**Development branch behavior for injected BBH event**

*Visits*

```
221.505 -50.503 i 1
223.318 -52.417 i 2
225.065 -49.449 i 3
226.201 -50.975 i 4
227.808 -51.995 i 5
223.318 -52.417 r 6
224.148 -52.630 r 7
224.851 -49.572 r 8
225.065 -49.449 r 9
227.808 -51.995 r 10
228.566 -51.050 r 11
223.025 -51.842 u 12
224.148 -52.630 u 13
224.851 -49.572 u 14
225.949 -49.293 u 15
227.967 -51.856 u 16
228.566 -51.050 u 17
```

![Dev-branch behavior](https://github.com/user-attachments/assets/b087d895-df20-4f6b-9036-3eb7ba6170ac)

**Main branch behavior for injected BBH event**

*Visits*

```
223.025 -51.842 i 1
224.148 -52.630 i 2
224.851 -49.572 i 3
225.949 -49.293 i 4
227.967 -51.856 i 5
228.566 -51.050 i 6
223.025 -51.842 r 7
224.148 -52.630 r 8
224.851 -49.572 r 9
225.949 -49.293 r 10
227.967 -51.856 r 11
228.566 -51.050 r 12
223.025 -51.842 u 13
224.148 -52.630 u 14
224.851 -49.572 u 15
225.949 -49.293 u 16
227.967 -51.856 u 17
228.566 -51.050 u 18
```

![Main branch behavior](https://github.com/user-attachments/assets/4fac1b34-4c4f-46b3-a54e-93cadbd52068)

** Summary **

Based on this simulation, I'm not quite sure that this addresses the intended issue. Additionally, the dev branch skips the first visit in i-band. 

It's interesting that we can't replicate the filter-visit behavior here - I tried a few other configurations with different visits and ToO footprints, to no avail. I wasn't using the sim runner here, I used a `CoreScheduler` object with the ToO survey from above added.

rhiannonlynne:
I think you're saying here,  that the PR should NOT be merged. 
I'm wondering if you looked at just adding a BandSortDetailer to the survey? This should also order the visits in filter order, while not changing anything else.

PR #206, opened 2025-07-24
Update too_scripted_surveys.py to have updated exposure times and visit numbers

seanmacb:
From a camera safety perspective, exposures longer than 30s (especially 120 and 180s) poses a significant risk of tripping HV on the camera, which has many cascading effects. 

For this reason, I revised the exposure times and visit numbers in the ToO surveys to ensure that the necessary depth is still achieved while keeping the Camera safe. 

To perform the image coaddition, work is needed on the image processing side. This is the route that is being employed by the bespoke image processing pipeline, and we will take the lessons learned from that process forward to the automated implementation.


yoachim (approved):
Looks all in order

PR #209, opened 2025-08-05
add update_conditions method to Survey objects.

yoachim:
Add `update_conditions` method to `Survey` objects so ToO surveys can generate list of desired observations before reward functions are called.

### DIFF
--- a/rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
+++ b/rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
@@ -451,16 +451,19 @@ class ToOScriptedSurvey(ScriptedSurvey, BaseMarkovSurvey):
                     observations = detailer(observations, conditions, target_o_o=target_o_o)
                 self.set_script(observations)
 
-    def calc_reward_function(self, conditions):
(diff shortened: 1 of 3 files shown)


rubin_scheduler/scheduler/surveys/too_scripted_surveys.py
```diff
@@ -451,17 +451,20 @@ def _new_event(self, target_o_o, conditions):
...
-        """If there is an observation ready to go, execute it,
-        otherwise, -inf"""
-        # check if any new event has come in
-
+    def update_conditions(self, conditions):
         if conditions.targets_of_opportunity is not None:
             for target_o_o in conditions.targets_of_opportunity:
                 if target_o_o.id > self.last_event_id:
                     self._new_event(target_o_o, conditions)
                     self.last_event_id = target_o_o.id
```
rhiannonlynne, line 459: I think this is all we have to work with currently, but I don't know how the alert packets (SCIMMA + too interpretation in lsst_ts + EFD) end up getting or keeping their numbering. 
We should check with @seanmacb

rhiannonlynne (approved):

PR #210, opened 2025-08-08
Add DateSwapBandScheduler

rhiannonlynne:
Summit schedules filter changes on particular dates, which prenight simulations need to match (and more detailed upcoming-weeks-ish simulations such as for SV survey) need to incorporate too, to help provide feedback on filter changes. 

If we want to simulate visits from the start of the survey, to check that predictions match actual outputs, this will also be necessary to synchronize filter changes (so should keep historical records too). 

This also gives us a way to provide limited filter set options during periods when a filter may be missing.


rhiannonlynne:
Hang on .. I realized that this is not quite the right setup. 
The previous (known) filter changes should be built-in and immutable, and then the upcoming changes should be modifiable via the kwarg without disturbing the known changes. 
I will go look up the history for lsstcam throughout SV and implement this.

PR #211, opened 2025-08-11
Make night be calculated rather than set by user

yoachim:
Have the `Conditions` object compute the `night` value rather than have it manually set. The `CoreScheduler` will override `conditions.survey_start_mjd` value with it's own.


tribeiro (commented):
Just a note that this is probably going to break backwards compatibility, as we are currently setting this on the CSC. So, would be good if you could provide backwards compatibility. Maybe provide a set method that warns of deprecation?

PR #212, opened 2025-08-11
dither detailers pre-calculate offsets for all dithers

yoachim:
Update dither detailers so offset positions are pre-computed, eliminating the risk that dither positions could be repeated if the scheduler is restarted each night.


rubin_scheduler/scheduler/detailers/dither_detailer.py
```diff
@@ -51,24 +52,42 @@ def __init__(self, max_dither=0.7, seed=42, per_night=True, nnights=7305):
         self.max_dither = np.radians(max_dither)
         self.per_night = per_night
         self.rng = np.random.default_rng(seed)
+
+        self.n_positions = copy.copy(nnights)
```
rhiannonlynne, line 56: Ok, this is for my own learning -- why do you need a copy here? 
Isn't `nnights` just an integer, so assigning self.n_positions to the integer would always be a copy? 
I'm assuming that the copy was needed because of the if not per_night clause below, but don't see why unless I'm not understanding what data type "nnights" is.

rubin_scheduler/scheduler/detailers/dither_detailer.py
```diff
@@ -51,24 +52,42 @@ def __init__(self, max_dither=0.7, seed=42, per_night=True, nnights=7305):
...
+
+        if not per_night:
+            nnights = (nnights, nnights)
```
rhiannonlynne, line 59: Just like previously .. the reason this is `(nnights, nnights)` is just because you wanted to make sure  that there were enough positions within a night? 
I still think it would be clearer if the second number was something other than `nnights` or if it at least had 
`# (number of nights, number of positions inside a night)` above as a comment line

There isn't anything technically wrong with this as it is, it's just that a new developer coming into this a few years down the road has no clue why this is the way it is, without going to the _generate_offsets method and puzzling through how things are set up and indexed (and there are no comment lines there either). 

self.n_positions  is likewise a little confusing, because I thought it was being used to indicate how many positions total there were for the dithers .. but it's not. It's only used in the per-visit dither offsets, to ensure that we cycle back to the start if we run over in a night (which would require a full 14 hour night of exposures shorter than 5s ..). 
Did you mean to maybe also include it in the per-night dither offsets as well?

tests/scheduler/test_detailers.py
```diff
@@ -425,6 +425,65 @@ def test_region_label(self):
...
+        # Now want a unique position
+        detailer = detailers.DitherDetailer(per_night=False)
+
+        offsets = detailer._generate_offsets(100, 2)
+        ra_a = offsets[:, 0]
+        dec_a = offsets[:, 1]
+
+        assert np.size(np.unique(ra_a)) == 100
+
+        # call again should result in different values
```
rhiannonlynne, line 480: I'm not sure I understand this -- if the dither detailer is called per_night=True, it pre-generates all of the offsets, so the values are fixed? But if you call it per_night=False, then it does not pre-generate all of the offsets, so they change? 
This is probably not unreasonable (I imagine you wanted to avoid saving all of the dithers for the entire survey, if they were per-visit). It should be documented however (there's no mention of this in the doc-string). 

The other thing I was wondering about was if the FBS was re-enabled during the night and doing per-visit dithers, this means the dithers would repeat within the night? Again also probably not terrible, but likely something to make sure is documented as expected behavior.

rhiannonlynne (approved):
Some comments inline, but I think all solvable with some documentation.

tests/scheduler/test_detailers.py
yoachim: It is always pre-generating positions. This is testing that `per_night` behaves properly and gives only one offset value if `True` and a unique position per visit if `False`.

Depends on what "re-enabled" means. If you re-instantiate the scheduler and play in completed observations, then we lose the count and potentially repeat dither positions. If you restore from a pickle, then the count is fine and there shouldn't be repeats.

rhiannonlynne: We talked about this on slack, but there were two issues: 
I didn't quite understand the API  for `_generate_offsets` and thought calling it with the same numbers meant the same values (but instead, internally this method increments a counter). 
The second issue is that an internal counter in a detailer works for simulations but probably doesn't work at the summit, since anything that happens as a result of `request_observations` is lost in the pickle/restore cycle. Only the results of `update_conditions` and `add_observation` are kept and updated in the FBS.

rhiannonlynne: Adding a `feature` has resolved the problem I think.

rhiannonlynne (approved):

rhiannonlynne (approved):
(approved but I added a comment about how 
"Note that if scheduler is re-build during a night .." 
is probably not correct due to playback of existing observations which would increment the feature counter again).

PR #214, opened 2025-08-14
tickets/SP-2447: fix bug in night computation

ehneilsen:


rhiannonlynne:
I don't think night being negative is necessarily a problem or bug. 
Making 'night' be single-value and corresponding to a particular night is better.
But then again - this ought to become calculated elsewhere shortly, so I'm not that fussed.

PR #215, opened 2025-08-14
set default survey start to be in the day

yoachim:
Roll survey start back 0.5 days to ensure sims don't start in the middle of the night.  

Add unit test to make sure SURVEY_START_MJD is during the day for the observatory.


ehneilsen (approved):
Looks good to me

PR #218, opened 2025-09-02
update AltAzShadowMaskBasisFunction to compute masks at intermediate …

yoachim:
Update `AltAzShadowMaskBasisFunction` so masks are computed at set `timesteps` to prevent gaps in the masks.


rhiannonlynne:
Can we do one more thing to make this more backwards compatible with old scheduler snapshots (just for a little while; can drop later):
 
 in the calc_reward, if self.time_step doesn't exist (like it would not. for an old snapshot) - just set it to something large? 
 Either that or something like this: 
 ```
        try:
            n_steps = np.floor(self.shadow_time / self.time_step)
        except AttributeError:
            n_steps = 0
```

rhiannonlynne:
Otherwise, looks good.

rhiannonlynne (approved):

PR #219, opened 2025-09-02
add shared numpy array tool

yoachim:
Handy utility for sharing numpy arrays when using multiprocessing.


ehneilsen (approved):

PR #223, opened 2025-09-15
U/yoachim/target name

yoachim:
* add some basis functions useful for gathering templates in good seeing conditions
* fix potential memory leak in dither detailer
* use lower case for ddf names

### DIFF
--- a/rubin_scheduler/scheduler/detailers/dither_detailer.py
+++ b/rubin_scheduler/scheduler/detailers/dither_detailer.py
@@ -403,7 +403,7 @@ class SplitDetailer(BaseDetailer):
 
 class CameraRotDetailer(BaseDetailer):
     """
-    Randomly set the camera rotation, either for each exposure, 
+    Randomly set the camera rotation, either for each exposure,
     for each time the detailer is called, or per night.
 
     Note dithering for each exposure may not be repeatable if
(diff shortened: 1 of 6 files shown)


rubin_scheduler/scheduler/detailers/dither_detailer.py
```diff
@@ -427,11 +431,15 @@ def __init__(
         dither="night",
         per_night=None,
         seed=42,
-        nnights=7305,
+        nnights=None,
         telescope="rubin",
+        n_in_night=7310,
```
rhiannonlynne, line 436: This default number continues to not make sense to me. This is the number of dither positions in a single night?

rubin_scheduler/scheduler/features/features.py
```diff
@@ -292,11 +292,16 @@ class NObservations(BaseSurveyFeature):
...
         Deprecated in favor of scheduler_note, but provided for backward
         compatibility. Will be removed in the future.
+    seeing_limit : `float`
+        The limit to apply on accepting image quality.
     """
 
-    def __init__(self, bandname=None, nside=DEFAULT_NSIDE, scheduler_note=None, survey_name=None):
+    def __init__(
+        self, bandname=None, nside=DEFAULT_NSIDE, scheduler_note=None, survey_name=None, seeing_limit=None
+    ):
```
rhiannonlynne, line 301: The unit test needs to be expanded to include testing the seeing limit.

rhiannonlynne:
Regarding the Nobs feature update (that then trickles up to the FootprintBasisFunction and MaskAfterNObsSeeingBasisFunction) -- I'm not arguing this doesn't make sense to do that this way, but I'm curious why you decided not to set a seeing limit for the *survey* and then have the survey intercept at "add_observations" and "add_observations_array" like they do for the "ignore_obs" strings?

rhiannonlynne (approved):

PR #224, opened 2025-09-17
add ecliptic_target util

yoachim:
Adding simple utility for setting region around the ecliptic.


rhiannonlynne:
One thought here - would it make more sense to put the "ecliptic_target" into the sky_area module, since it doesn't (by itself) have anything to do with footprints and is a static sky area definition? 
Generally, if I was going to do this randomly on the fly for example, I would call 
`rubin_scheduler.utils.sky_area.generate_all_sky` and use the returned map. 

I realize that you don't need all of the converted coordinates (ecliptic and galactic) that are returned in the all sky map dictionary, so ecliptic_target is more efficient, but it's basically the same idea with slightly different constraints (max dec is defined by the max_dec value, instead of defining min and max declination via the min_alt value), so maybe is more natural in the same module.

ehneilsen (approved):

PR #226, opened 2025-09-18
add max to camera footprint

yoachim:
Have the camera footprint mask off anything beyond a given radius. Default of 1.94 degrees is a pretty minor change compared to just the full focal plane

### DIFF
--- a/rubin_scheduler/utils/camera_footprint.py
+++ b/rubin_scheduler/utils/camera_footprint.py
@@ -21,6 +21,9 @@ def rotate(x, y, rotation_angle_rad):
 class LsstCameraFootprint:
     """
     Identify point on the sky within an LSST camera footprint.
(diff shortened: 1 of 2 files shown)


rhiannonlynne:
It would be a good idea to make a comment in the docstring about why this is in place. 
Also, that 1.94deg = 350mm in the focal plane (and that 1.75 deg =317mm in the focal plane, which is the actual requirement .. but I think Eli is saying he can get to 1.94).

rhiannonlynne:
I'd say also link to whatever tech note or whatever Eli has made, but I'm not quite sure where it is .. if you can dig it up that might be helpful documentation as well for when we later come back and say what the heck. 
(there. might be something in an LCR?).

rhiannonlynne (approved):

PR #227, opened 2025-09-23
Tickets/sp 2621

yoachim:
Detailer that should help if pre-scheduled DDF observations get backed up and need to execute before morning twilight.

### DIFF
--- a/rubin_scheduler/scheduler/surveys/scripted_surveys.py
+++ b/rubin_scheduler/scheduler/surveys/scripted_surveys.py
@@ -149,7 +149,7 @@ class ScriptedSurvey(BaseSurvey):
     def calc_reward_function(self, conditions):
         """If there is an observation ready to go, execute it,
         otherwise, -inf"""
-        observation = self.generate_observations_rough(conditions)
+        observation = self.generate_observations(conditions)
         if (observation is None) | (np.size(observation) == 0):
             self.reward = -np.inf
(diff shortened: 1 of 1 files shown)


rubin_scheduler/scheduler/detailers/detailer.py
```diff
@@ -202,6 +203,55 @@ def __call__(self, observation_array, conditions):
...
+        Estimate of filter change time
+    visit_overhead : `float`
+        Estimate of visit overhead (read, slew, settle).
+        Default 2.5 (seconds).
+    twilight_cut_to : `str`
+        Cut to sun altitude of -18 degrees or -12 degrees.
+        Valid values of "n18" or "n12".
+    """
+
+    def __init__(self, pad=5.0, filter_change_time=120.0, visit_overhead=2.5, twilight_cut_to="n18"):
```
rhiannonlynne, line 225: We're slewing between DDF visits within a night now, right? Isn't 2.5 too short? (this is already shorter than the readout time).

rhiannonlynne (approved):

rubin_scheduler/scheduler/detailers/detailer.py
yoachim: Sure, we can bump that up. With the 5-minute pad default it shouldn't end up mattering in most situations.

PR #269, opened 2026-03-24
add filter match for scripted surveys

yoachim:
Add filter match for scripted surveys. Was already in `add_observations_array` method, so should have been here to match already.


rhiannonlynne (commented):
Does this work for the ToO surveys where sometimes the filter is swapped by a detailer?

rhiannonlynne:
I'm a little confused - does this still need the PR? Is it just because of the docstring update, or is the other change meaningful?

rhiannonlynne (approved):

PR #270, opened 2026-03-25
add BlobPairSurvey

yoachim:
Blob survey that can be useful for taking pairs with dithers in between.


yoachim:
Doing some more changes

PR #274, opened 2026-04-09
SP-3063: split wind basis function to mask and regular bf

yoachim:
Split `AvoidDirectWind` into a wind pressure masking basis function and a wind pressure preference basis function. Refactored to put the calculation of wind pressure in the `Conditions` object.

### DIFF
--- a/rubin_scheduler/scheduler/basis_functions/basis_functions.py
+++ b/rubin_scheduler/scheduler/basis_functions/basis_functions.py
@@ -1396,18 +1396,35 @@ class WindPressureBasisFunction(BaseBasisFunction):
 
     Parameters
     ----------
+    wind_speed_maximum : `float`, optional
+        Wind speed to mark regions as unobservable (in m/s).
+    wind_speed_minimum : `float`, optional
+        If wind speed blow this level, return no effective reward
+        map. Default 0 (m/s).
     nside : `int`, optional
         The nside for the basis function. Default None uses
         `set_default_nside()`.
     """
 
-    def __init__(self, nside=DEFAULT_NSIDE):
+    def __init__(self, wind_speed_maximum=20.0, wind_speed_minimum=0.0, nside=DEFAULT_NSIDE):
+
+        if wind_speed_maximum < wind_speed_minimum:
+            raise ValueError(
+                "wind_speed_maximum of %f m/s is less than wind_speed_minimum value of %f m/s."
+                % (wind_speed_maximum, wind_speed_minimum)
+            )
         super().__init__(nside=nside)
+        self.wind_speed_maximum = wind_speed_maximum
+        self.wind_speed_minimum = wind_speed_minimum
 
     def _calc_value(self, conditions, indx=None):
         reward_map = np.zeros(hp.nside2npix(self.nside))
 
-        if conditions.wind_speed is None or conditions.wind_direction is None:
+        if (
+            conditions.wind_speed is None
+            or conditions.wind_direction is None
+            or conditions.wind_speed < self.wind_speed_minimum
+        ):
             return reward_map
 
         reward_map -= conditions.wind_pressure**2.0
--- a/rubin_scheduler/scheduler/basis_functions/basis_functions.py
+++ b/rubin_scheduler/scheduler/basis_functions/basis_functions.py
@@ -1399,7 +1399,7 @@ class WindPressureBasisFunction(BaseBasisFunction):
     wind_speed_maximum : `float`, optional
         Wind speed to mark regions as unobservable (in m/s).
     wind_speed_minimum : `float`, optional
-        If wind speed blow this level, return no effective reward
+        If wind speed below this level, return no effective reward
         map. Default 0 (m/s).
     nside : `int`, optional
         The nside for the basis function. Default None uses
--- a/rubin_scheduler/scheduler/basis_functions/basis_functions.py
+++ b/rubin_scheduler/scheduler/basis_functions/basis_functions.py
@@ -33,6 +33,7 @@ __all__ = (
     "FilterDistBasisFunction",
     "RewardRisingBasisFunction",
     "send_unused_deprecation_warning",
+    "WindPressureBasisFunction",
 )
 
 import warnings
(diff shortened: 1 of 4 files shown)


rubin_scheduler/scheduler/basis_functions/basis_functions.py
```diff
@@ -1388,6 +1391,30 @@ def _calc_value(self, conditions, indx=None):
...
+    def __init__(self, nside=DEFAULT_NSIDE):
+        super().__init__(nside=nside)
+
+    def _calc_value(self, conditions, indx=None):
+        reward_map = np.zeros(hp.nside2npix(self.nside))
+
+        if conditions.wind_speed is None or conditions.wind_direction is None:
+            return reward_map
+
+        reward_map -= conditions.wind_pressure**2.0
```
rhiannonlynne, line 1430: For this - we should probably use Elana's modifications. 
In other words (basically),  do not penalize looking into the wind. Maybe dependent on wind speed.
I'm not quite sure how to use wind speed now the conditions and the BF uses wind pressure,  but probably there is a straightforward enough calculation of max wind pressure that can be used instead of wind speed directly.

rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
```diff
@@ -27,6 +28,35 @@
 from rubin_scheduler.utils import DEFAULT_NSIDE, SURVEY_START_MJD, _angular_separation, _hp_grow_mask
 
 
+class MaskDirectWindBasisFunction(BaseBasisFunction):
+    """Mask area looking into the wind
```
rhiannonlynne: Mask areas where the wind pressure is a level determined by the defined wind_speed_maximum.  (?) 
(it's not just mask looking into the wind I thought).

rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
```diff
@@ -27,6 +28,35 @@
...
+        self.wind_speed_maximum = wind_speed_maximum
+
+    def _calc_value(self, conditions, indx=None):
+        reward_map = np.zeros(hp.nside2npix(self.nside))
+
+        if conditions.wind_speed is None or conditions.wind_direction is None:
+            return reward_map
+
+        mask = conditions.wind_pressure > self.wind_speed_maximum
+        reward_map[mask] = np.nan
```
rhiannonlynne, line 55: Using the calculated wind pressure to mask based on a wind speed value always felt really wrong to me. Wind pressure and wind speed do not have the same units. 
Can we fix this?

rhiannonlynne (changes requested):
This is going the right way, but should be aligned with the recent changes requested by Elana and maybe we can finally fix the wind pressure != wind speed issue.

rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
yoachim: Our wind pressure is wind speed time a cos term, so I think they do have the same units.

rhiannonlynne: I think the problem is that what is called "wind pressure" is not a pressure (force/area) but rather something like the wind speed perpendicular to the TMA azimuth.

rhiannonlynne:
@elanaku @tribeiro 
Elana, does this align with your desired update to the wind basis function? 
 have split the wind functions into 
 * a simple mask/0 set of values ("MaskDirectWindBasisFunction") which just masks off areas where the wind speed perpendicular to the TMA azimuth is more than the wind_speed_maximum value (i.e. masks into and out of the wind, once it's above the max windspeed). I think at the moment, this would be used for AuxTel with some max windspeed below the closing speed (??) and for Simonyi  either with a large max wind speed, such that we'd probably close the telescope first or not used at all (we have no evidence for wind shake with Simonyi, is my understanding). 
 * a basis function that generates a reward that is proportional to the negative value of wind speed perpendicular to the TMA squared .. because it is negative this is "reward" acts a penalty, encouraging observing at +/- 90 degrees to the wind direction. However, it does include the cutoff for wind speed minimum, below which there is no preference. 
 So -- this still does not actively encourage observing toward the wind, under any circumstances. 
 
 I believe, for Simonyi, we would leave the reward basis function out of the survey configuration until we understand better the static behaviour of the seeing as a function of wind speed and direction .. We had also talked about simply including this information into the seeing model, if this is understood as a static behaviour of the observatory as a function of wind. 
 
 We have not yet really gotten enough information about how the airflow in the dome needs to behave if we need to use the telescope pointing to more actively modify the airflow for flushing hot air or encouraging more laminar flow (potentially, for example, at the start of the night). I think that's likely to need something more complicated than either of these basis functions (i.e. something like the reward function, but that also knows about the temperature in and out of the dome maybe, or at least the time of night).

rubin_scheduler/scheduler/basis_functions/basis_functions.py
```diff
@@ -1388,6 +1391,47 @@ def _calc_value(self, conditions, indx=None):
         return reward_map
 
 
+class WindPressureBasisFunction(BaseBasisFunction):
+    """Have negaitive reward for wind pressure squared.
```
rhiannonlynne, line 1395: Set a negative reward based on the effective wind speed squared.

rhiannonlynne (approved):
If you wouldn't mind updating the last docstring slightly (can remove the typo, but also clarify the effect). 

I was trying to decide if we care that the reward is now discontinuous wind / no-wind, but I guess if the basis function is applied to all surveys in a tier, then we probably don't.

rubin_scheduler/scheduler/basis_functions/basis_functions.py
```diff
@@ -1388,6 +1391,47 @@ def _calc_value(self, conditions, indx=None):
...
+
+    Parameters
+    ----------
+    wind_speed_maximum : `float`, optional
+        Wind speed to mark regions as unobservable (in m/s).
+    wind_speed_minimum : `float`, optional
+        If wind speed blow this level, return no effective reward
```
elanaku: I think this is a small typo. Could it say 'If wind speed **below** this level..."?

elanaku (commented):
Looks great!

PR #275, opened 2026-04-13
no default target_name if not set

yoachim:
Auto-generating a target_name forces a warrning if anyone tries to set their own target_name via a detailer.

### DIFF
--- a/rubin_scheduler/scheduler/surveys/field_survey.py
+++ b/rubin_scheduler/scheduler/surveys/field_survey.py
@@ -106,9 +106,6 @@ class FieldSurvey(BaseSurvey):
         self.survey_name = survey_name
         if self.survey_name is None:
             self._generate_survey_name(target_name=target_name)
-        # Backfill target name if it wasn't set
-        if target_name is None:
-            target_name = self.survey_name
 
         super().__init__(
             nside=nside,


rhiannonlynne:
I'm not sure I understand your comment in the PR. 
Were you trying to say that "Currently, automatically setting the target name in the FieldSurvey triggers a warning message if someone set target_name via a detailer.  This change allows a user to set the target_name in the Survey, but does not set a value if there is None set - letting the detailer work without a warning. "
??

rhiannonlynne:
Did this actually come up as an issue in a configuration somewhere? 
I'm thinking that this PR is fine, I'm not sure I see the problem it was solving, but that it probably means we should go check the configurations active in ts_config_scheduler to ensure they're all setting the target_name as they intend (it was old-practice to just use survey_name and assume it would be carried through .. none of those configurations used a detailer to set the target_name).

yoachim:
I think this came up in the Roman config--that has a `LabelRegionsAndDDFs`, so it always throws a warning that the `target_name` is also set, even though it was only automatically set.

yoachim:
looking in `ts_fbs_utils`, all the calls to `FieldScheduler` have a `target_name` set except for the roman config. So this should be safe and not change anything unexpectedly downstream.

rhiannonlynne (approved):

PR #277, opened 2026-04-14
U/yoachim/seeing limit array fix

yoachim:
Lingering spots in `add_observation_array` methods that need to be updated for declination dependent seeing limit. Will add a unit test to loop over all features to double check `add_observation` and `add_observation_array` work as expected.

### DIFF
--- a/rubin_scheduler/scheduler/features/features.py
+++ b/rubin_scheduler/scheduler/features/features.py
@@ -334,7 +334,9 @@ class NObservations(BaseSurveyFeature):
 
             if np.size(self.seeing_limit) > 1:
                 valid_indx[
-                    np.where(IntRounded(seeing_vals) >= IntRounded(self.seeing_limit[observations_hpid]))[0]
+                    np.where(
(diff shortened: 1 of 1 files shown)


rhiannonlynne:
Thanks - I forgot about it, but I thought I had run across something a problem that I thought was related to this. Unfortunately, I failed to note where/how it happened, so hopefully this "just fixes it".

rhiannonlynne (approved):

rhiannonlynne:
When you're working on the unit test -- please check that add_observation_array works with multiple observations but also with only a single observation? 
I had an error adding the ongoing observations via add_observation_array one night, and while I didn't investigate, the error message reported made me think that it was probably because there was only one observation that fell into the relevant "bucket" for some feature or other (I think it was in the near-sun twilight survey). 
As more observations were acquired, the problem went away, so I do think it was probably related to something not quite working when the length of the array was only 1 (or possibly 0, after some cut or other).

PR #278, opened 2026-04-20
SP-3091: address queue manager masking more than required pixels; extend capabilities for current config (for templates and dynapairs)

yoachim:
* Update the QueueManager so masked pixels don't grow. 
* add kward to `MaskAfterNObsSeeingBasisFunction` so it can be continuous or reset per season
* Update scheduler_note on blob surveys so it records how many observations are being attempted in a blob (so it's easy to see how much it may have expanded/contracted blob from ideal pair time)

### DIFF
--- a/rubin_scheduler/scheduler/schedulers/queue_manager.py
+++ b/rubin_scheduler/scheduler/schedulers/queue_manager.py
@@ -7,6 +7,7 @@ import numpy as np
 
 from rubin_scheduler.scheduler.detailers import RotspUpdateDetailer
 from rubin_scheduler.scheduler.utils import IntRounded, ObservationArray
+from rubin_scheduler.utils import _ra_dec2_hpid
 
 
 class BaseQueueManager:
@@ -134,12 +135,17 @@ class BaseQueueManager:
         # need to interpolate to the actual positions we want.
         # now to interpolate to the reward positions
         if np.size(reward) > 1:
-            reward_interp = hp.get_interp_val(
-                reward,
-                np.degrees(self.desired_observations_array["RA"][valid]),
-                np.degrees(self.desired_observations_array["dec"][valid]),
-                lonlat=True,
+            # hp.get_interp_val uses 4 nearest neighbors, so that can
+            # go nan if a farther helpix is masked. This uses the
+            # HEALpix the RA,dec falls in. Technically should probably be
+            # the nearest neighbor HEALpix center.
+            nside = hp.npix2nside(reward.size)
+            hpids = _ra_dec2_hpid(
+                nside,
+                self.desired_observations_array["RA"][valid],
+                self.desired_observations_array["dec"][valid],
             )
+            reward_interp = reward[hpids]
             sub_valid_reward = np.isfinite(reward_interp)
             valid = np.ravel(valid[sub_valid_reward])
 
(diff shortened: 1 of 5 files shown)


rubin_scheduler/scheduler/schedulers/queue_manager.py
```diff
@@ -134,12 +135,17 @@ def find_valid_observations_indx(self, conditions, mjd=None):
...
-                lonlat=True,
+            # hp.get_interp_val uses 4 nearest neighbors, so that can
+            # go nan if a farther helpix is masked. This uses the
+            # HEALpix the RA,dec falls in. Technically should probably be
+            # the nearest neighbor HEALpix center.
+            nside = hp.npix2nside(reward.size)
+            hpids = _ra_dec2_hpid(
+                nside,
+                self.desired_observations_array["RA"][valid],
+                self.desired_observations_array["dec"][valid],
```
rhiannonlynne, line 146: Is this actually a single hpid not multiple hpid or is it multiple? 
I am thinking that it's multiple, because there are multiple observations. 

I am starting to wonder if it makes sense for the `observation` array to carry the hpid of the original reward map .. 
we go from reward map -> hpids with the best reward -> add the tessellation to get the ra/dec of the boresight .. and then I think we might go backwards to hpid in multiple places? 
(am I imagining that we do this ra_dec_2_hpid in multiple places?)

rhiannonlynne (commented):
I wish this was split into separate PRs. 
The MaskAfterNObsSeeingBasisFunction is straightforward  and approved. 

I don't understand quite what you're doing with the queue manager and the interpolation. A given observation only has one RA/Dec value, so should only return one hpid, shouldn't it? So then why do you need to interpolate the result? 

And I would like to see an example of what the scheduler note looks like after the changes around the dynapairs. 
I'm also not sure the docstring and kwargs make logical sense -- the scaling is not just for twilight, is it? Doesn't it also scale when there are approaching prescheduled observations? Should the kwargs be updated to be more clear as to the purpose of the scaling?

yoachim:
Previous scheduler note:  `pair_15, iz, a`

Updated scheduler note: `pair_33, iz, bs 52, a`

("bs" short for "blob size" I guess? Just trying to keep character count down).

yoachim:
The `BlobSurvey` has kwargs `twilight_scale` to tell it to scale up to twilight, and then `check_scheduled` to tell it to scale up to any scheduled observations. And then `in_twilight` if one is trying to complete blobs before twilight ends.

rubin_scheduler/scheduler/schedulers/queue_manager.py
yoachim: This is now returning a single hpid per observation. Previously, I just used the built-in healpy interpolation to get a mask value at each RA,dec. This caused problems because healpy grabs the 4 closest healpix centers and interpolates, so if any of the 4 closest points are masked it returns nan, causing masks to grow unexpectedly. 

I don't think we can do hpid in `observation_array` in a useful way because each observation overlaps multiple healpix, so it can be ambiguous which healpix drove the selection of a pointing. 

We probably have an ra_dec_2_hpid call in the scripted surveys as well. I'm not too worried about doing that call twice--it's an array operation and I think healpy is using something smart to do it. It's not a killer like full coordinate conversions of ra,dec to alt,az that have to call multiple trig functions.

rhiannonlynne (approved):

PR #279, opened 2026-04-22
tickets/SP-3114: Accept str as well as EarthLocation in riseset_times, and use a str default

ehneilsen:
If EarthLocation is used as the default, of_site gets run on import, and this can cause the import to fail if the astropy database of site locations is down.


yoachim (approved):
yay!

rhiannonlynne (approved):

PR #280, opened 2026-04-22
extend unit test for Features

yoachim:


ehneilsen (approved):
Not clear on what this change is for specifically, but it looks fine to me.

PR #282, opened 2026-05-26
SP-3167: Handle special cases of add_observations_array

rhiannonlynne:
If the length of the observations_array ends up being zero due to ignoring all incoming observations, some features can crash -- so this PR adds a check first. 

In addition, if add_observations_array is handed a length 1 array, things can fail in finding the healpix ids, so fall back to just add_observation. 

These are edge-cases, but I found them when doing the following: 
* set up a CoreScheduler with all observations prior to the current night
* add on-going observations so far using add_observations_array
* if the first observations of the night are only twilight_near_earth visits, many other surveys ignore them .. and then features which look for the '.max()' of one of the values will crash because the array is 0 length by that time.

### DIFF
--- a/rubin_scheduler/scheduler/schedulers/core_scheduler.py
+++ b/rubin_scheduler/scheduler/schedulers/core_scheduler.py
@@ -160,6 +160,10 @@ class CoreScheduler:
         the observation array and constructed observations + healpix id
         to each survey.
         """
+        if len(obs) == 1:
+            self.add_observation(obs)
+            return
+
         # Need to add "band" here if it wasn't populated
         missing_band = np.where(obs["band"] == "")
         band = np.char.rstrip(obs["filter"][missing_band], chars="_0123456789")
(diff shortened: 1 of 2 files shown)


rubin_scheduler/scheduler/schedulers/core_scheduler.py
```diff
@@ -160,6 +160,10 @@ def add_observations_array(self, obs):
         the observation array and constructed observations + healpix id
         to each survey.
         """
+        if len(obs) == 1:
```
yoachim, line 163: Seems like this should never trip because in our usual configuration a single observation would always overlap more than 1 HEALpix. Should it throw a warning that a 1-element array was passed and the user probably wants to use the `add_observation` method instead?

yoachim (approved):

rubin_scheduler/scheduler/schedulers/core_scheduler.py
rhiannonlynne: I was able to trip this if I passed add_observation_array a single observation though, and it triggered an actual Error. 
I don't think a warning in that case would be useful, because it would have already crashed the higher-level process. 

Perhaps a different option though would be: 
try -- the usual
except Exception as e -- report the exception and fall back to adding observations one at a time.

(the point being, Tiago has moved to add_observation_array at the summit for startup, thank goodness, but if that crashes then the SchedulerCSC goes down and observing has to stop .. it's unlikely that he'll ever run add_observation_array with a single observation, but anything can happen).

PR #283, opened 2026-05-29
Tickets/sp 3181

yoachim:
* Move bulk of altAzShadow computation to a method on the `Conditions` object so it can be cached
* Update the ignore_obs checking to use faster numpy array rather than list comprehension
* Refactor of survey `add_observation` method to reduce repeat code in subclasses

Results in 2.2x speedup.

### DIFF
--- a/rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
+++ b/rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
@@ -460,6 +460,11 @@ class AltAzShadowMaskBasisFunction(BaseBasisFunction):
 
     def _calc_value(self, conditions, indx=None):
 
+        # Computation of the actual alt/az mask is done in the
+        # conditions object this lets the result of the computation
+        # be cached and shared among the surveys which have
+        # the same kwargs.
+
         result = conditions.alt_az_shadow_mask(
             conditions.mjd,
             self.nside,
--- a/rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
+++ b/rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
@@ -25,7 +25,7 @@ import numpy as np
 import rubin_scheduler.scheduler.features as features
 from rubin_scheduler.scheduler.basis_functions import BaseBasisFunction
 from rubin_scheduler.scheduler.utils import CurrentAreaMap, HpInLsstFov, IntRounded
-from rubin_scheduler.utils import DEFAULT_NSIDE, SURVEY_START_MJD, _angular_separation, _hp_grow_mask
+from rubin_scheduler.utils import DEFAULT_NSIDE, SURVEY_START_MJD, _angular_separation
 
 
 class MaskDirectWindBasisFunction(BaseBasisFunction):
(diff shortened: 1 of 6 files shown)


rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
```diff
@@ -459,110 +459,19 @@ def __init__(
...
+            self.nside,
+            min_alt=self.min_alt,
+            max_alt=self.max_alt,
+            min_az=self.min_az,
+            max_az=self.max_az,
+            shadow_time=self.shadow_time,
+            pad=self.pad,
+            scale=self.scale,
+            time_step=self.time_step,
+        )
```
rhiannonlynne, line 479: This is caching for members of the same time, mostly right? 
The only thing I dislike about having this attached to the conditions is that it's a lot of computation to put into the Conditions object and it's a bit more opaque there. 
Maybe put a bit of a comment line in here then to explain why it's there instead of here? This will help us understand the code in the future. 

I was thinking something along the lines of 

```
# computation of the actual alt/az mask is done in the conditions object
# this lets the result of the computation be cached and shared among 
# the surveys which have the same general conditions for the alt/az shadow mask
# In general, this is likely all members of the same tier as they typically
# share the alt/az limits and shadow_time.
```

rubin_scheduler/scheduler/features/conditions.py
```diff
@@ -580,6 +581,152 @@ def az_to_antisun(self):
             self.calc_az_to_antisun()
         return self._az_to_antisun
 
+    @functools.lru_cache(maxsize=5)
```
rhiannonlynne, line 586: Did the number 5 come out of the profiler? 
I imagine this is best set to something like the number of distinct alt/az/shadow_mask_time combinations we have (since the time_step is typically linked to the shadow_time and the rest (except for MJD) are all pretty standard things we don't change). 
It might be worth setting a comment here as well, so that in the future you don't have to just write "magic number" above the 5.

rubin_scheduler/scheduler/surveys/base_survey.py
```diff
@@ -217,21 +218,40 @@ def add_observations_array(self, observations_array_in, observations_hpid_in):
...
-                self.extra_features[feature].add_observation(observation, **kwargs)
-            for bf in self.extra_basis_functions:
-                self.extra_basis_functions[bf].add_observation(observation, **kwargs)
-            for bf in self.basis_functions:
-                bf.add_observation(observation, **kwargs)
-            for detailer in self.detailers:
-                detailer.add_observation(observation, **kwargs)
-            self.reward_checked = False
+        if len(self.ignore_obs_array) > 0:
+            sub_str_indx = np.strings.find(observation["scheduler_note"][0], self.ignore_obs_array)
```
rhiannonlynne, line 235: Can you add a clarifying comment here as well, because I am never certain if "match the note" means "exactly match the note" or "a substring of the scheduler note matches the 'ignore' string".

rubin_scheduler/scheduler/surveys/base_survey.py
```diff
@@ -217,21 +218,40 @@ def add_observations_array(self, observations_array_in, observations_hpid_in):
...
+        """
+        for feature in self.extra_features:
+            self.extra_features[feature].add_observation(observation, **kwargs)
+        for bf in self.extra_basis_functions:
+            self.extra_basis_functions[bf].add_observation(observation, **kwargs)
+        for bf in self.basis_functions:
+            bf.add_observation(observation, **kwargs)
+        for detailer in self.detailers:
+            detailer.add_observation(observation, **kwargs)
+        self.reward_checked = False
```
rhiannonlynne, line 256: I'm not quite clear why you have added add_loops here but not add_loops_array

If it's to implify code (removing the add_loops in the ScriptedSurvey) then you should probably also have an add_loops_array so that you can also remove those same kinds of lines here https://github.com/lsst/rubin_scheduler/blob/bb2d8f4a2bfe86974671573182f11681b764902e/rubin_scheduler/scheduler/surveys/scripted_surveys.py#L126

(I think mostly here my brain is objecting that you made "add_observation" more complicated and different looking than add_observation_array, which makes it more complicated to read and understand how they are the same/different down the road).

rubin_scheduler/scheduler/surveys/base_survey.py
rhiannonlynne: Also maybe "add_loops" isn't quite the right name? 
maybe 
_propagate_observation
_distribute_observation
_register_observation 
??

rhiannonlynne:
This looks good, but I do have some minor requests that I think will help with maintainability and clarity down the road. 
Do you know which one of these changes made the biggest difference in the overall timing?

rhiannonlynne (changes requested):

rubin_scheduler/scheduler/basis_functions/mask_basis_funcs.py
yoachim: I wanted to make it a stand-along function, but it requires so many things that are in the conditions object that going stand-alone not all the kwargs were hashable. 

Note added.

rubin_scheduler/scheduler/surveys/base_survey.py
yoachim: Sure. Let's go with `sub_objects_add_observation`

rubin_scheduler/scheduler/surveys/base_survey.py
yoachim: Yeah, I wasn't touching `add_observation_array`, but it's easy enough to keep it analogous to `add_observation`.

rubin_scheduler/scheduler/features/conditions.py
yoachim: Note added.

rhiannonlynne (approved):

PR #285, opened 2026-06-08
Tickets/sp 3218

yoachim:
New kwarg so `CoreScheduler` will flush its queue if a new ToO event is passed in via `update_conditions`. No checking if the ToO is actually visible, so it's up to the user to confirm they think the ToO is worth flushing the queue for before sending it in. Or, if the user is fine interrupting something like pairs to be sure the ToO doesn't have observations that can be done immediately.

### DIFF
--- a/rubin_scheduler/scheduler/features/conditions.py
+++ b/rubin_scheduler/scheduler/features/conditions.py
@@ -906,8 +906,8 @@ class Conditions:
         moon_phase_sunset : `float`
             The phase of the moon (0-100 illuminated) at sunset.
             Useful for setting which bands should be loaded.
-        targets_of_opportunity : `list` [`rubin_scheduler.scheduler.targetoO`]
-            targetoO objects.
+        targets_of_opportunity : `list` [`TargetoO`]
+            TargetoO objects like `rubin_scheduler.scheduler.utils.TargetoO`.
         planet_positions : `dict` {`str`: `float`}
             Dictionary of planet name and coordinate e.g., 'venus_RA',
             'mars_dec'
--- a/rubin_scheduler/scheduler/schedulers/core_scheduler.py
+++ b/rubin_scheduler/scheduler/schedulers/core_scheduler.py
@@ -254,12 +254,15 @@ class CoreScheduler:
         if self.flush_for_new_too:
             if conditions_in.targets_of_opportunity is not None:
                 if len(conditions_in.targets_of_opportunity) > 0:
-                    current_ids = set([too.id for too in conditions_in.targets_of_opportunity])
-                    new_ids = current_ids.difference(self.too_alert_ids_set)
-                    if len(new_ids) > 0:
-                        self.flush_queue()
-                        for new_id in new_ids:
-                            self.too_alert_ids_set.add(new_id)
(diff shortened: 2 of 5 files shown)


rhiannonlynne:
Is there any way we can make this smarter, so that it does NOT flush the queue for a ToO which is not yet visible? This would happen often I would think. 
Presumably we will not get triggers for ToOs which will not be worth observing. This is the job of the ToO board and the upstream scimma + vetting services. However, we may indeed get triggers for things like GWs which are not visible from Rubin yet. 

Maybe we should talk about the logic, because I think there must be something I didn't anticipate. 
ToOs we receive which do not need immediate followup will reserve "room" for themselves via their scheduled_obs lists + "respect scheduled_obs" basis functions, is my understanding. 
And ToOs we receive which could receive immediate followup would do the same thing, but the earliest they could observe (without this new kwarg) would be when the currently active queue is finished. 

How could we make things work so that it's more like: 
* if there is a (only new?) ToO which could take observations right now, but hasn't been respected by the scheduled_obs, that it could take priority?   
In the past, we only got updated Conditions (and thus ToO triggers) when the queue was already empty. Now that we get updates every 2 observations, the queue is not necessarily empty. I imagine this is the difference that can lead to scheduled_obs not being respected (which could even happen if it was just that time had gone by while in a fault state).

yoachim:
Sounds like you want to be able to send in all the ToO events as they happen, and have the scheduler decide if 
1) it should interrupt what it's doing (flush the queue so ToO observation can start immediately), or 
2) if it should just pass the ToO event along so the ToO visits get generated, but finish whatever is in the queue at the moment

One solution would be to add a method to `TargetoO` that has the logic for if the ToO is urgent and (probably) visible. Could be some edge cases like we flush a queue for a ToO that's right next to the moon, but would mostly eliminate unnecessary queue flushes.

rhiannonlynne:
Ok, before we do that work let's just ask Tiago and Sean about what the plan is for populating the too events themselves. Maybe I'm wrong and they're planning to hold onto upcoming events until they're visible. I just wouldn't have thought they'd know quite enough at the event generation level to know about the moon etc.

rhiannonlynne:
All right - Sean confirmed the idea is to just send all events as they are available. 
Your update seems like primarily it will evaluate if the altitude of the ToO is above a given limit (either the centre of the ToO or the any part of the footprint) and if so, flush the queue. 
This seems reasonable enough to me. The only second thing I would wonder is for large footprint things, would it be wise to wait until a significant part of the footprint is above the alt limit? But also -- I am guessing that's a reasonable starting point and we can update more later. 

How much have you tested this in simulations of a night or two with a ToO event arriving both above and below and just at the horizon?

yoachim:
Now there's a flag so we can fully simulate ToOs interrupting. Generating and populating extra `Conditions` objects slows things down quite a bit, so it's not set as the default.

rhiannonlynne:
I think that will be needed to evaluate the impact of real clouds as well. 

I take it things worked as expected?

yoachim:
Yup, it does something slightly different when turning the flag on, so the ToOs are interrupting and making the queue flush.

rhiannonlynne (approved):

PR #286, opened 2026-06-29
fix and unit tests for _set_block_size method

yoachim:
Fix repeated variable name that was causing issues expanding/contracting block sizes properly if there were both scheduled observations and approaching twilight.


rhiannonlynne (approved):

rhiannonlynne:
Did you already look at the effect on sim?

yoachim:
Yup. Makes it scale the pair length more times like it should

PR #287, opened 2026-07-05
SP-3269: make duration a kwarg with a default

rhiannonlynne:
Made TargetooO  'duration' default to None instead of being a mandatory argument. SchedulerCSC does not want to fill this number for ToO (and the duration is not relevant to FBS execution itself, only the SimTooServer, which gets filled by the summit in operations .. we need to match summit, but will probably need other machinery).  Tiago requested that it be changed so he does not have to send a value here.
Using "None" is somewhat arbitrary, but is distinguishable from real float values set on purpose. Note that SimToOServer fills in 3 at present, for cases of None so even if not set, will work as it does at present in simulations.
The duration is useful for simulations so is still available to use (even if sim ToOs all seem to use 3 at present).

Moved TargetofO and simTargetooServer to their own files, to make them easier to find. 

I added unit tests for both TargetoO and simTargetooServer. 
In doing so, the fact that nan pixels in the footprint could be the "np.max" value and still not evaluate == 0 came up .. which would have left "queue flush" = True instead of flipping it to False. I added some extra code to handle this case. (the doctstring says np.nan can be used for do-not-observe pixels).

### DIFF
--- a/rubin_scheduler/scheduler/utils/__init__.py
+++ b/rubin_scheduler/scheduler/utils/__init__.py
@@ -2,5 +2,6 @@ from .comcam_tessellate import *
 from .footprints import *
 from .observation_array import *
 from .sky_area import *
+from .too_objects import *
 from .tsp import *
 from .utils import *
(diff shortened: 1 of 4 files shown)


yoachim (approved):
Looks good

PR #289, opened 2026-07-08
feasibility basis function for not running when cloudy

yoachim:
Add a feasibility function for preventing a survey from executing when it's cloudy.

### DIFF
--- a/rubin_scheduler/scheduler/basis_functions/feasibility_funcs.py
+++ b/rubin_scheduler/scheduler/basis_functions/feasibility_funcs.py
@@ -696,7 +696,7 @@ class CloudedOutMapBasisFunction(BaseBasisFunction):
     say things are not feasible.
     """
 
-    def __init__(self, median_cloud_limit=0.1):
+    def __init__(self, median_cloud_limit=1.0):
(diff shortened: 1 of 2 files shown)


rhiannonlynne:
What's the use-case for stopping a survey based on the average cloud extinction, instead of the extinction map and the pointings desired for the given survey?

yoachim:
You said we should turn off the template tier if it's cloudy.

rhiannonlynne:
Ok - just for template tier. 
I'm not sure if we decided the whole sky got brighter or just some of it .. but probably good to have the option here regardless. 

I guess we don't yet get the uncertainty, although it might be nice to be able to update this when we do so that the median doesn't overweight parts of the map which are more uncertain. 

Can you make the default larger though?  Maybe 1 mag?  I don't really know what DM will use as the cutoff transparency value, but something between 0.5 and 1 mag might be closer, and I am hoping they can go more like 1 mag so we can still use more visits.

rhiannonlynne:
Ah - and it probably ought to be a np.nanmedian because I think nans can get into the cloud map data.

rhiannonlynne (approved):

PR #293, opened 2026-07-30
Tickets/sp 3310

yoachim:
Add a new `ModelObservatory` subclass so we can set the observatory to observe at sun altitudes higher than -12 degrees.

### DIFF
--- a/rubin_scheduler/scheduler/model_observatory/bright_observatory.py
+++ b/rubin_scheduler/scheduler/model_observatory/bright_observatory.py
@@ -1,0 +1,95 @@
+__all__ = ("BrightObservatoryModel",)
+
+from .model_observatory import ModelObservatory
+from rubin_scheduler.utils import NextTimeSun
+from astropy.time import Time
+from rubin_scheduler.utils import SURVEY_START_MJD
+
+
+class BrightObservatoryModel(ModelObservatory):
+    """Observatory that can go beyond sun altitude of -12 deg.
+
+    Parameters
+    ----------
+    sun_rise_limit_deg : `float`
+        The sun altitude limit where we should skip ahead
+        to the next sunset. Default -9 (degrees).
+    sun_set_limit_deg : `float`
+        What should the altitude of the sun be at sunset
+        when observing starts. Default -9 (degrees).
+    delta_time_step_sec : `float`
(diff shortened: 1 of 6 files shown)


rubin_scheduler/scheduler/model_observatory/bright_observatory.py
```diff
@@ -0,0 +1,146 @@
...
+
+import numpy as np
+from astropy.time import Time
+
+from rubin_scheduler.utils import SURVEY_START_MJD, NextTimeSun, _angular_separation
+
+from .model_observatory import ModelObservatory
+
+
+class BrightObservatoryModel(ModelObservatory):
```
ehneilsen, line 11: It's not obvious that having a separate class is the right way to do this: it seems like the "normal" ModelObservatory is just a (not very) special case of this class, which is the more general. If anything, maybe have the more general class (currently called BrightObservatoryModel) be the base class and the normal instance (ModelObservatory) be the subclass.

rubin_scheduler/utils/sun_position.py
```diff
@@ -0,0 +1,96 @@
...
+
+import numpy as np
+from astropy.coordinates import AltAz, EarthLocation, get_body, get_sun
+from astropy.time import Time
+from scipy.optimize import minimize
+
+from rubin_scheduler.utils import Site
+
+
+class NextTimeSun(object):
```
ehneilsen, line 11: I think this is redundant with `rubin_scheduler.utils.riseset`.

ehneilsen (approved):
Comments are all minor and optional, but I think worth considering.

rubin_scheduler/scheduler/model_observatory/bright_observatory.py
yoachim: I think the best answer would be to have an even more general base class, maybe that's not as site specific, then `ModelObservatory` and `BrightObservatoryModel` could be sublcasses of that. I'll put in a ticket to refactor that way, but it might not be urgent.

rubin_scheduler/utils/sun_position.py
yoachim: I'll put in a ticket to speed test and replace with `riseset` if there's no penalty.

